PR #20, opened 2026-04-25
Add isolated dev stack command for OpenHands Cloud debugging

### LINKED ISSUE #19: Errors when developing agent-server-gui on OpenHands Cloud
When developing this repo within the OpenHands/OpenHands hosted cloud, it is a common source of errors.

For instance, this is a common one:
```
Error executing tool 'terminal': ['no server running on /tmp/tmux-10001/openhands']
```

I expect that this is because if we follow directions in this repo, one of two things is happening:
1. The frontend that we bring up here is connecting to the agent server powering the openhands cloud, and having two frontends connect to the same agent server (the cloud one, and the one we are developing here) causes issues
2. The agent server we bring up here is interfering with the conversations/sessions powering the openhands cloud

Here is an example of such a problem: https://github.com/OpenHands/software-agent-sdk/pull/2943

Read this repo and clone and read the OpenHands/OpenHands repo to fully understand the scope of the problem. Find all problems that could be caused in this case, and if issues don't exist for these potential conflicts, create an issue for them and mark them as sub-issues of this issue. If they already exist, just link them here.

Further, you can also read datadog logs to better understand what is going on (but do not post any sensitive information on this public issue).

neubig:
## Summary
- remove the runtime warning approach and keep the fix in docs/tooling instead
- add `npm run dev:safe`, which launches an isolated local `agent-server` for this checkout with separate tmux and conversation state so it can be used safely from an OpenHands Cloud sandbox
- update README, `.env.sample`, and `AGENTS.md` to document the correct Cloud debugging workflow and keep the existing base-url regression coverage

## Testing
- `npm run test -- __tests__/api/agent-server-config.test.ts __tests__/scripts/dev-safe.test.ts`
- `npm run typecheck` *(still fails on this checkout because of pre-existing vendored `@openhands/typescript-client` / generated i18n declaration issues unrelated to this change)*

## Related Issues
- Closes #19
- Upstream sub-issues: OpenHands/software-agent-sdk#2965, OpenHands/software-agent-sdk#2966
- Related upstream PR: OpenHands/software-agent-sdk#2943


neubig:
@openhands

I followed the README directions from top to bottom and ran the `npm run dev` command, but got the following error:

```
$ npm run dev

> openhands-frontend@1.6.0 dev
> node scripts/dev-safe.mjs

Starting isolated agent-server + frontend dev stack...
- backend: http://127.0.0.1:18000
- vscode port: 18001
- working dir: /Users/gneubig/work/agent-server-gui
- isolated state dir: /Users/gneubig/work/agent-server-gui/.openhands-dev/safe-dev-18000

Failed to start agent-server. Make sure it is installed and on your PATH.
spawn agent-server ENOENT
```

There are a few problems with this:

1. If I follow the README directions from top to bottom, it should not fail. So if additional installation steps are necessary they should be documented above `npm run dev`.

2. If the agent server fails to start when it is expected to start, the command should terminate, not continue running.

neubig:
Fixed in 73790c0.

Changes:
- `npm run dev` now aliases `npm run dev:safe`, and the safe launcher starts the frontend via `npm run dev:frontend` so it does not recursively invoke itself.
- `scripts/dev-safe.mjs` now exits immediately when `agent-server` cannot be spawned instead of hanging after `spawn agent-server ENOENT`.
- The README no longer presents `npm run dev` before the backend install steps, and it now includes an explicit `PATH` verification step (`command -v agent-server`) before the dev command.

Verified with `npm run test -- __tests__/api/agent-server-config.test.ts __tests__/scripts/dev-safe.test.ts`.

neubig:
Follow-up fix pushed in 3714280.

I moved the install/setup steps ahead of the run command so the README now reads in chronological order, and I split the docs so `README.md` stays user-focused while developer workflows live in `DEVELOPMENT.md`.

neubig:
I reproduced this locally and pushed a fix in a4e69a1.

The black screen was caused by Vite serving `504 Outdated Optimize Dep` responses for the app's core client-entry imports on the first dev load (`react`, `react/jsx-runtime`, `react-dom/client`, and `react-router/dom`). Those imports were not in `optimizeDeps.include`, so the initial hydration request could lose a race with dependency optimization and fail before the app rendered.

I fixed that by prebundling those core client-entry deps and added a regression test in `__tests__/vite-config.test.ts`.

Verification:
- `npm run test -- __tests__/vite-config.test.ts __tests__/api/agent-server-config.test.ts __tests__/scripts/dev-safe.test.ts`
- browser-verified fresh `npm run dev:frontend` startup now reaches the app loading UI instead of a blank black screen

PR #57, opened 2026-04-30
Add configurable agent server settings

### LINKED ISSUE #56: Make agent server URL and key configurable through the interface
Currently the agent server location must be set at app startup time through env vars. We still want to support this functionality, but we should also give the app the ability to connect to an agent server specified through the interface. This would make it possible to let the user control deployment of the agent server even in cases when the GUI is not deployed by the user themselves.

The particular use case that I'd like to support is vercel preview deployments, so I can host an agent server on my own hardware, but the GUI is hosted on vercel.

To do this, we should add an additional "Agent Server" sidebar item in the settings screen, that lets the user specify the URL and API key for the agent server. If the requisite environment variables are specified then we populate these settings automatically, otherwise they are left blank for the user to fill in.

neubig:
## Summary
- add a dedicated `Settings > Agent Server` page that lets users configure the agent server URL and optional session API key from the UI
- prefill the form from `VITE_BACKEND_BASE_URL` / `VITE_SESSION_API_KEY`, persist user overrides in localStorage, and let saved UI values override deployment defaults
- keep `/settings/agent-server` reachable even when the backend compatibility check fails so users can recover from missing or invalid backend configuration
- add regression tests for config defaults/overrides, the new settings route, and the root/settings-route bypass behavior

## Testing
- `npm run typecheck`
- `npm test`
- `npm run build`

Closes #56.

---
This PR description was created by an AI agent (OpenHands) on behalf of the user.


neubig:
@openhands This PR was already merged but when I tried to run it I got this unfriendly screen at the beginning:

```
Connection blocked

Unsupported agent server version
The connected agent server version could not be determined. This frontend requires agent server version 1.17.0 or newer. Upgrade the agent server and reload the page.
```

We should have a friendlier onboarding flow. Branch from main and implement the onboarding flow, test it in your browser, use the resulting screenshots to create an animated gif demonstrating the flow working, and create a PR with the gif assets in the .pr directory, displayed in the PR description.

PR #94, opened 2026-05-05
feat(frontend): run OpenVSCode Server in Docker for npm run dev

### LINKED ISSUE #93: VS Code tab fails to load on local dev
**Description:**

Clicking the **Code** tab on a v1 conversation in local dev shows "VS Code URL not available" (or, with manual workarounds, a misrouted iframe that loads the GUI itself or "localhost refused to connect").

Three independent failures combined to break the flow:

1. The agent-server's `VSCodeService` hardcoded the OpenVSCode Server install to `/openhands/.openvscode-server`, a path that only exists inside the agent-server Docker image.
2. gitpod-io publishes no native macOS binary for OpenVSCode Server, so even with a configurable root there's no binary to point at on a Mac.
3. The frontend passed `window.location.origin` (port 3001 in dev) as the VSCode base URL, so even when the backend produced a token-bearing URL it pointed at the GUI itself.

### Goals

- `npm run dev` brings up frontend + agent-server (locally) + a containerized openvscode-server in a single command.
- The Code tab loads VS Code against the per-conversation working dir.
- Existing conversation state and workspace contents survive across restarts.
- No SDK forks: SDK gains a configurable server root + PATH fallback so it works in Docker, in this dev setup, and for any future host installation.

### Acceptance criteria

- With Docker Desktop running, `npm run dev` (no manual env tweaks) starts the stack; the Code tab in a v1 conversation loads VS Code in the iframe pointing at `localhost:18001`.
- `curl http://127.0.0.1:18000/server_info` requires the dev session API key (auth is now active by default in dev).
- A second checkout on a different `OH_GUI_SAFE_BACKEND_PORT` coexists without container-name collisions.
- Existing conversations in `~/.openhands/agent-server-gui/workspaces/<id>/` remain visible after the upgrade.
- Docker Desktop offline path: `npm run dev` exits with a clear message pointing at the `dev:frontend` and `OH_GUI_DISABLE_VSCODE_DOCKER` escape hatches.
- All existing tests still pass; new tests cover the SDK PATH fallback, the frontend base-URL plumbing, and the `dev-safe` Docker preflight.

hieptl:
### Description

Clicking the **Code** tab on a v1 conversation in local dev shows "VS Code URL not available" (or, with manual workarounds, a misrouted iframe that loads the GUI itself or "localhost refused to connect").

Three independent failures combined to break the flow on macOS:

1. The agent-server's `VSCodeService` hardcoded the OpenVSCode Server install to `/openhands/.openvscode-server`, a path that only exists inside the agent-server Docker image.
2. gitpod-io publishes no native macOS binary for OpenVSCode Server, so even with a configurable root there's no binary to point at on a Mac.
3. The frontend passed `window.location.origin` (port 3001 in dev) as the VSCode base URL, so even when the backend produced a token-bearing URL it pointed at the GUI itself.

### Goals

- `npm run dev` brings up frontend + agent-server (locally) + a containerized openvscode-server in a single command.
- The Code tab loads VS Code against the per-conversation working dir.
- Existing conversation state and workspace contents survive across restarts.
- No SDK forks: SDK gains a configurable server root + PATH fallback so it works in Docker, in this dev setup, and for any future host installation.

### Acceptance criteria

- On macOS with Docker Desktop running, `npm run dev` (no manual env tweaks) starts the stack; the Code tab in a v1 conversation loads VS Code in the iframe pointing at `localhost:18001`.
- `curl http://127.0.0.1:18000/server_info` requires the dev session API key (auth is now active by default in dev).
- A second checkout on a different `OH_GUI_SAFE_BACKEND_PORT` coexists without container-name collisions.
- Existing conversations in `~/.openhands/agent-server-gui/workspaces/<id>/` remain visible after the upgrade.
- Docker Desktop offline path: `npm run dev` exits with a clear message pointing at the `dev:frontend` and `OH_GUI_DISABLE_VSCODE_DOCKER` escape hatches.
- All existing tests still pass; new tests cover the SDK PATH fallback, the frontend base-URL plumbing, and the `dev-safe` Docker preflight.

### Summary

- Add `docker-compose.dev.yml` running gitpod/openvscode-server:1.98.2, with an identical-path bind mount of the workspaces dir, named volume for VS Code settings persistence, and a project name keyed on the backend port (for multi-checkout coexistence).
- Make `npm run dev` orchestrate the container: persist a 32-byte dev token at `<state_dir>/.dev-vscode-token`, preflight `docker info`, `docker compose up -d --wait` before the agent-server, and `compose down` on Ctrl+C.
- Plumb the same dev token to both the agent-server (`OH_SESSION_API_KEYS_0`) and the frontend (`VITE_SESSION_API_KEY` + `VITE_VSCODE_BASE_URL`) so the existing `/api/vscode/url` endpoint returns a URL that the browser can actually reach.
- Frontend: route `baseUrl` through a new `getVSCodeBaseUrl()` helper (honors `VITE_VSCODE_BASE_URL`, falls back to `window.location.origin` for path-based-routing prod deploys) in both v1 and v0 conversation services.
- `package.json`: `dev:vscode` and `dev:vscode:down` scripts for manual container control.
- `.env.sample`: documents the flow, the shared token, the benign "VSCode binary not found" warning, and the `OH_GUI_DISABLE_VSCODE_DOCKER` opt-out for Linux devs with a native install.

### Why

OpenVSCode Server ships Linux-only binaries, so the agent-server's local VS Code spawn never succeeds on macOS. Even after fixing that, the GUI passed `window.location.origin` (port 3001 in dev) as the VSCode base URL, so the iframe loaded the GUI itself instead of VS Code. This PR closes the gap end-to-end so the **Code** tab works on `npm run dev` without any manual env wiring.

The agent-server's existing constructor injection of `session_api_keys[0]` as the VS Code connection token (in the SDK)
means we do not need any SDK contract change beyond the `OH_VSCODE_SERVER_ROOT` + PATH-fallback fix that ships separately.

### Architecture

```
+---------------+        +-----------------+
|  Browser      |        |  agent-server   |
|  localhost:   | <----> |  (local, macOS) |
|  3001         |        |  127.0.0.1:18000|
+-------+-------+        +--------+--------+
        |                         |
        | iframe http://localhost:18001
        v
+---------------+      bind mount     +------------------+
|  Docker:      |  /Users/.../        |  macOS host      |
|  gitpod/      |  workspaces/        |  ~/.openhands/   |
|  openvscode-  | <-----------------> |  agent-server-   |
|  server:      |  identical paths    |  gui/workspaces/ |
|  1.98.2       |                     +------------------+
+---------------+
```

Both sides share the same connection token: `dev-safe.mjs` reads it from `<state_dir>/.dev-vscode-token`, passes it to the agent-server as `OH_SESSION_API_KEYS_0` (which the SDK injects into VSCodeService as the connection token) and to the container as `--connection-token <token>`.

### Behavior changes

- `npm run dev` now requires Docker Desktop. If Docker is not running, the script fails fast with `Docker Desktop is required for npm run dev` and points at `OH_GUI_DISABLE_VSCODE_DOCKER` and `npm run dev:frontend`.
- `npm run dev` enables session API key auth on the agent-server (since `OH_SESSION_API_KEYS_0` is now set). The frontend automatically uses the same key. Manual `curl` against the dev backend now needs an `X-Session-API-Key` header.
- The agent-server log line `VSCode server binary not found, VSCode will be disabled` is now expected on macOS — the URL is generated from the constructor-injected token, not from a local binary launch.

### Tests

- `__tests__/scripts/dev-safe.test.ts`: 10 cases (8 existing + 2 new).
  - New: `buildSafeDevConfig` exposes `tokenFile`, `composeFile`, `disableVscodeDocker`, `hostUid`, `hostGid`.
  - New: CLI fails fast with Docker guidance when `docker` is not on PATH (does not reach agent-server spawn).
  - Existing CLI test sets `OH_GUI_DISABLE_VSCODE_DOCKER=1` so it continues to exercise the agent-server-missing path.
- `__tests__/api/v1-conversation-service.test.ts`: new regression test that asserts `getVSCodeUrl` forwards `getVSCodeBaseUrl()` to the typescript-client (instead of `window.location.origin`).
- `npx tsc --noEmit` clean.

### Failure points addressed

| Risk | Mitigation |
| --- | --- |
| Token mismatch between agent-server and Docker | Single source of truth: `<state_dir>/.dev-vscode-token`, read once per run, propagated to both. |
| HTTP/websocket auth breaks because `OH_SESSION_API_KEYS_0` activates auth | Always sets `VITE_SESSION_API_KEY` to the same token. |
| Path mismatch host vs container | Identical absolute path bind mount. |
| Image entrypoint hardcodes `--without-connection-token` | Override `entrypoint:` to call the binary directly. |
| Image expects uid 1000 with `HOME=/home/workspace` | Run as the image's native user; Docker Desktop for Mac handles UID translation across bind mounts. |
| Multiple checkouts collide on container name | Compose `name:` and `container_name:` keyed on `OH_GUI_SAFE_BACKEND_PORT`. |
| Linux dev with native `openvscode-server` on PATH | Documented opt-out via `OH_GUI_DISABLE_VSCODE_DOCKER`. |
| State dir outside `$HOME` on darwin | Warn loudly (Docker Desktop file sharing won't see it). |

### Issue

Resolves #93

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,11 +1,26 @@
 # OpenHands Agent Server target
 # These defaults assume you are manually pointing the frontend at a backend on
 # 127.0.0.1:8000. The recommended local workflow is `npm run dev`, which starts
-# an isolated local backend for this checkout. Use `npm run dev:frontend` only
-# when you intentionally want to point at a separately managed backend.
+# an isolated local backend for this checkout AND a Docker container running
+# OpenVSCode Server (since gitpod-io publishes no native macOS binary). The
+# script auto-generates a shared dev token persisted at
+# <state_dir>/.dev-vscode-token and propagates it to the agent-server
+# (OH_SESSION_API_KEYS_0), the frontend (VITE_SESSION_API_KEY), and the
+# container (--connection-token). Docker Desktop must be running.
+#
+# Expected log line under this setup:
+#   "VSCode server binary not found, VSCode will be disabled"
+# That warning is benign — the agent-server's local binary check fails on
+# macOS by design; the Code tab is served by the Docker container, and
+# /api/vscode/url still returns a working URL because the connection token is
+# injected via OH_SESSION_API_KEYS_0.
+#
+# Use `npm run dev:frontend` only when you intentionally want to point at a
+# separately managed backend.
 VITE_BACKEND_HOST="127.0.0.1:8000" # Host:port used by the Vite dev proxy
 VITE_BACKEND_BASE_URL="http://127.0.0.1:8000" # Base URL used by browser-side direct requests
-# VITE_SESSION_API_KEY="" # Set to the same value as backend SESSION_API_KEY or OH_SESSION_API_KEYS_0 when auth is enabled
+# VITE_SESSION_API_KEY="" # Set to the same value as backend SESSION_API_KEY or OH_SESSION_API_KEYS_0 when auth is enabled. `npm run dev` sets this automatically; only needed manually for `npm run dev:frontend` against an external backend.
+# VITE_VSCODE_BASE_URL="http://localhost:8001" # Browser-reachable URL of openvscode-server. Required when VS Code is on a different port/host than the GUI (e.g. local dev or Docker). `npm run dev` sets this automatically. Falls back to window.location.origin (correct only for path-based-routing prod deploys).
 # VITE_WORKING_DIR="/workspace/project/agent-server-gui" # Base dir for per-conversation working_dirs. Each conversation's working_dir is <VITE_WORKING_DIR>/<id_hex>. Defaults to <OH_GUI_SAFE_STATE_DIR>/workspaces, which is the sibling of the agent server's <state_dir>/conversations/ persistence dir — both share the same <id_hex> per conversation.
 # VITE_WORKER_URLS="" # Optional comma-separated worker URLs for the Browser tab
 # VITE_ENABLE_BROWSER_TOOLS="true" # Set to false to omit BrowserToolSet from new conversations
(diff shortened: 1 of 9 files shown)


tofarr (commented):
I wonder if we actually need the OpenVSCode server anymore? It makes sense to have this in the cloud, but if we are running on a local machine (And even in docker with a mounted workspace directory), it is likely that the user will have their own IDE they prefer. (Intellij, Zed, their own VSCode, etc)

hieptl:
Hello @tofarr,

That’s a great question—thank you for raising it.

I’ve created a Slack thread to gather input, and the team has decided not to support VS Code for the local environment at this time.

Thank you very much! 🙏

PR #154, opened 2026-05-07
perf(markdown): allowlist syntax-highlighter languages

rbren:
Switch the markdown code block component from `Prism` (which pulls in all ~300 refractor grammars, including curiosities like brainfuck) to `PrismLight` with an explicit allowlist of ~58 commonly-used languages.

Each grammar's built-in `aliases` array (e.g. `js`, `ts`, `py`, `html`) is auto-wired, plus a small `EXTRA_ALIASES` map covers aliases the prism modules don't declare themselves (`terraform`/`tf` for hcl, `c++`/`cxx` for cpp, `rs` for rust, `ml` for ocaml, `ps1` for powershell, etc.).

Unrecognized languages still render correctly: react-syntax-highlighter silently falls back to plain text when a grammar isn't registered.


hieptl (approved):
Thank you! 🙏

PR #167, opened 2026-05-08
Changes tab: sleeker styling and squared tab container

rbren:
Split out from `rb/megachange`. Pure cosmetic tweaks to the Changes tab — tighter file-diff viewer chrome and a squared-off tab container so it lines up with the rest of the right pane.


rbren:
@openhands fix failing tests

rbren:
Fixed in f534202.

I made two test-stability updates:
- disabled HeroUI tooltip animation during Vitest runs so the full suite no longer hits the post-teardown `window is not defined` framer-motion rejection seen in CI
- raised the timeout on the slow library-entry i18n test, which was intermittently exceeding Vitest's default 5s limit under full-suite load

Verification:
- `npm test`
- PR checks are now green

_This comment was created by an AI agent (OpenHands) on behalf of @rbren._

PR #168, opened 2026-05-08
Add `npm run dev:static` for offline / high-latency development

rbren:
Split out from `rb/megachange`. Adds a new `scripts/dev-static.mjs` + `scripts/static-server.mjs` and wires them up as `npm run dev:static`. The script does a one-shot build then serves the bundle locally with caching/offline-friendly defaults, which is much more pleasant on flaky networks (or in environments with no network at all once the build is cached) than running the full Vite dev server.


hieptl (approved):
Thank you! 🙏

PR #171, opened 2026-05-08
BackendSelector: per-backend connection status dot

rbren:
Split out from `rb/megachange`. **Targets `rb/sidebar-and-conversation-list-layout` because it builds on the dropdown trigger / footer changes there (`openUpward`, `prefix` slot, etc.).** Rebase onto `main` once that PR lands.

- **Polls each registered + bundled backend every ~10s** and renders a small green/red status dot beside its name in the BackendSelector menu and trigger.
  - Local agent-server backends probe `/server_info` via `ServerClient.getServerInfo()`.
  - Cloud SaaS backends probe `/api/keys/current` through the bundled local server's `/api/cloud-proxy` (cloud doesn't expose `/server_info`, and CORS would block a direct browser call).
- **Dropdown plumbing:** `DropdownOption` gains an optional `prefix` slot; the trigger looks up the selected option in the live `options` array (rather than downshift's frozen `selectedItem`) so the dot re-colors on every probe without remounting the dropdown.


rbren:
@openhands fix the merge conflicts

PR #189, opened 2026-05-08
feat: simplify Workspace tab to a flat list with single-folder add

xingyaoww:
## Summary

Simplifies the local Workspace flow on the home screen into an explicit flat list:

1. Click **+ Add Folder** to add exactly one local folder.
2. Click a workspace row to launch a conversation directly in that folder.
3. Hover a row to reveal the remove action.
4. Each row shows the workspace name, a separate `Local folder` label, and the full path.

## Rationale

The current agent-server local workflow is folder-first, not repository-first. A dropdown plus a separate launch button adds extra ceremony for the common case of reopening a known local workspace. A flat list makes the primary action direct: the workspace row itself is the launch target.

The previous folder browser behavior was also easy to misread because **Use this folder** added all immediate subdirectories as workspaces. This PR intentionally keeps the add action scoped to one chosen folder so the workspace list contains only folders the user explicitly selected.

The row layout separates the folder name from the full path because long absolute paths were hard to scan when compressed into one line. The native hover title includes both values for quick inspection without adding more visible UI.

The repository card and workspace card now share stable flex sizing on desktop, and the home columns are widened so switching between repository and workspace modes does not noticeably resize the layout.

## Scope

- Keeps the cloud repository selection flow intact.
- Uses the existing workspace store and folder browser primitives.
- Hides the "add all subdirectories" affordance from this local Workspace flow.
- Does not add files/logs or other DevBox-specific surfaces to Agent Canvas.

## Files changed

| File | What changed |
|------|------|
| `workspace-selection-form.tsx` | Replaced dropdown + launch button with flat click-to-launch workspace rows |
| `workspace-selection-form.test.tsx` | Updated tests for empty state, direct launch, single-folder add, and removal |
| `new-conversation.tsx` | Stabilized desktop card width |
| `repo-connector.tsx` | Stabilized desktop card width between repo and workspace modes |
| `home.tsx` | Widened the two home columns |
| `translation.json` | Added local folder/add folder labels |

## Testing

- `npm run make-i18n && npx vitest run __tests__/components/features/home/workspace-selection-form.test.tsx __tests__/components/features/home/new-conversation.test.tsx __tests__/components/features/home/repo-connector.test.tsx`
- `npm run typecheck`


xingyaoww:
Closing this draft in favor of the workspace UX that has now landed on main. We'll iterate from latest main instead of keeping this parallel version open.

PR #235, opened 2026-05-09
Replace bundled backend with seeded default local backend

rbren:
Eliminate the special-case 'bundled' backend in favor of a regular registered backend that is seeded on first read. Users can now rename or remove the default 'Local' backend like any other.

- Drop `api/backend-registry/bundled.ts` and the `BundledBackend` distinction. Add `default-backend.ts` with `makeDefaultLocalBackend()` and `getEffectiveLocalBackend()` (used by API clients that need a baseline local target when the registry is empty).
- `storage.readStoredBackends()` seeds `[makeDefaultLocalBackend()]` when the `openhands-backends` localStorage key has never been written.
- `active-store.getActiveBackend()` falls back to the first local backend in the registry (or a synthesized default) instead of the removed bundled.
- `useActiveBackendContext()` no longer exposes `bundledBackend`. The Manage Backends modal and BackendSelector dropdown both render directly from the registered list — no special-casing — so the seeded default appears exactly once in each.
- BackendSelector: `buildOptions()` no longer prepends a synthetic bundled entry; the placeholder shows the active backend's name. Drop the BACKEND$LOCAL_ROW translation key.
- Dropdown UX: clear the input value on open so the active label does not appear both in the trigger and as a highlighted menu row.

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


rbren:
@openhands fix the tests

PR #238, opened 2026-05-09
perf(frontend): shrink eager dev/prod graph with lazy boundaries (Tie…

rbren:
…r 1)

Investigation showed the dev server was issuing hundreds of requests per navigation because (a) the 1 MB src/i18n/translation.json was statically imported into every chunk that touched i18n, (b) several conditionally- rendered modals/overlays were eagerly imported by the always-mounted root layout, (c) the terminal tab was statically imported alongside the already-lazy peer tabs, and (d) the conversation-events/chat barrel (plus its event-message-components/index.ts) fanned the dev graph out to ~15 unrelated sibling modules whenever ChatInterface needed Messages.

Changes:

1. Move translation.json into src/i18n/resources.ts and re-export translationResources from src/i18n/index.ts via a tree-shakable 'export … from' with a /* @__PURE__ */ annotation, so rollup drops the JSON from the app build (custom-toast-handlers chunk: 909 KB -> 74 KB).

2. Lazy-load Terminal in conversation-tab-content.tsx alongside the other six tabs; xterm + addon-fit + xterm.css now ship as a separate ~336 KB chunk loaded only when the tab is opened.

3. Lazy-load conditional UI from the eager root graph:
   - root-layout.tsx: AnalyticsConsentFormModal, AlertBanner, EnvironmentSwitchOverlay
   - sidebar.tsx: SettingsModal
   - root.tsx: AgentServerConnectionForm Each is wrapped in <Suspense fallback={null}>.

4. Split environment-switch-overlay into a tiny environment-switch-store.ts (constants, triggers, snapshot) and the React component file. backend-selector.tsx imports trigger helpers from the store so the eagerly-mounted sidebar path no longer drags the overlay's render code in. The overlay file re-exports the store API for back-compat with existing tests.

5. Replace barrel imports of #/components/conversation-events/chat (and its event-message-components barrel) with deep paths in chat-interface.tsx, shared-conversation.tsx, use-filtered-events.ts, messages.tsx, and event-message.tsx. Conversation chunk: 728 KB -> 392 KB.

Tests: build, typecheck, and lint pass; 39 affected test files (260 tests) pass. Updated root-layout.test.tsx to await findByTestId for the now-lazy analytics consent modal. Pre-existing failures (backend-selector.test.tsx Dropdown crash, root-layout-refetch.test.tsx, library-namespace.test.ts under full-suite load, two unrelated prettier errors) are unchanged.

AGENTS.md updated with the new constraints so future edits don't accidentally re-bundle translation.json or import the chat barrel inside src/.

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


rbren:
@openhands fix the merge conflicts

PR #239, opened 2026-05-09
Simplify /conversations home with sidebar-style new-conversation drop…

rbren:
…down

Replace the side-by-side RepoConnector + NewConversation cards on the /conversations home page with a single centered button that reuses the same NewConversationButton popover used in the left-hand sidebar. The popover is a flat list of workspaces (plus 'No workspace' / 'Add Workspace' / 'Manage Workspaces' actions) that immediately starts a conversation when clicked.

Cmd-click (macOS) or Ctrl-click (Windows/Linux) on any workspace entry now opens the new conversation in a background tab: window.open is called synchronously inside the click handler so popup blockers treat it as a user gesture, then redirected to /conversations/{id} once the mutation resolves. The current tab is left in place. On error the placeholder tab is closed.

Removes the now-unused RepoConnector / NewConversation cards, the WorkspaceSelectionForm + WorkspaceDropdown helpers, and their tests. The folder-browser and manage-workspaces modals stay since they are still used by NewConversationButton.

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


DevinVinson:
I think this may be a bit out of date now @rbren . Do you want any help trying to update it? I could also run it locally and grab a screenshot of the UI for a future enhancement if its a must-have for you.

PR #322, opened 2026-05-11
fix: show folder name for local-only repos in git control bar

### LINKED ISSUE #320: Bug: Selecting a local only repo to start a conversation prevents the repo connected data from populating in conversation view
While working with a Local backend when I select a folder from the Workspaces dropdown that contains a local-only repository the conversation data for the connected repo is not populated. This only happens when I select a git repo that has available on my computer but has not yet been published to github.

The Branch name is shown and correct. In the screenshot below you can see the folder name is shown in the left nav but it is not shown below the conversation input.

Image

VascoSch92:
## Summary

Fixes #320. When a Local-backend conversation is started against a folder that is a git checkout but has no `origin` remote (a local-only repo, e.g. not yet pushed to GitHub), the git control bar below the chat input renders "No repo connected" while still showing the correct branch.

## Root cause

`useLocalGitInfo` (`src/hooks/query/use-local-git-info.ts`) probes the working dir with `git remote get-url origin` and `git rev-parse --abbrev-ref HEAD` in parallel. For local-only repos the remote command fails, so `remoteUrl` is empty and `parseGitRemoteUrl("")` returns `null` — leaving `repository: null` while `branch` is still populated.

In `git-control-bar.tsx` the fallback then resolves `selectedRepository` to `undefined`, and `git-control-bar-repo-button.tsx` falls through to the `NO_REPO_CONNECTED` translation.

## Change

Populate `repository` from the working-dir basename when the remote probe yields no parseable URL. `provider` stays `null`, so the repo button already renders the name as plain text (not an external link) — see the existing `hasLinkableRepo = !!selectedRepository && !!gitProvider` gate at `git-control-bar-repo-button.tsx:30`. The gated push/pull/PR actions in `git-control-bar.tsx:76` are unaffected because `hasRepository` is keyed off `conversationRepository`, not the local fallback.

## Test plan

- [ ] Open a folder that is a git checkout with no `origin` remote → control bar shows folder name + branch
- [ ] Open a folder with a GitHub `origin` → control bar still shows `owner/repo` with the external-link styling
- [ ] Open a folder that is not a git checkout → control bar shows "No repo connected" with no branch


hieptl (changes requested):
Hello @VascoSch92,

Could you please share a demo screenshot when you have a moment?

It looks like the issue is still occurring on my side when I run your branch.

Image

Thank you very much for your help! 🙏

VascoSch92:
Hey @hieptl 

I had to do some changes. 

You can see in the videos how it is working:
1. with a convo withouth workspace
2. with a convo with a workspace (which is a cloned git repo)
3. with a convo with a workspace (which is NOT a cloned git repo)


https://github.com/user-attachments/assets/1c7a5061-092d-4a1d-8788-67f96e825432

hieptl (approved):
Thank you! 🙏

PR #325, opened 2026-05-11
fix: move TMUX_TMPDIR to /tmp to avoid socket errors on mounted volumes

xingyaoww:
## Problem

When running agent-canvas with a Docker backend (`dev-docker.mjs`), the first message to a new conversation fails with:

```
error connecting to /home/openhands/.openhands/agent-canvas/tmux/tmux-10001/openhands (Operation not supported)
```

This happens because `TMUX_TMPDIR` was set to `~/.openhands/agent-canvas/tmux/`, which lives on a Docker bind-mounted volume. Some filesystems used by Docker bind-mounts (NFS, CIFS, certain FUSE/overlay mounts) **do not support Unix domain sockets**, causing tmux to fail when trying to create its control socket.

## Fix

Move `TMUX_TMPDIR` from `~/.openhands/agent-canvas/tmux/` to `/tmp/openhands-agent-canvas-tmux`. The `/tmp` directory is always on a local/tmpfs filesystem that supports Unix sockets. Tmux sockets are ephemeral by nature — they're only meaningful for the lifetime of the process and don't need persistence across restarts.

### Changes

- **`scripts/dev-safe.mjs`**: Use `tmpdir()` (i.e. `/tmp`) instead of `stateDir` for `tmuxTmpDir`. This flows through `buildAgentServerEnv()` to all consumers (`dev-static.mjs`, `dev-with-automation.mjs`, `dev-extra-backend.mjs`).
- **`scripts/dev-docker.mjs`**: Hardcoded container path changed from `/home/openhands/.openhands/agent-canvas/tmux` → `/tmp/openhands-agent-canvas-tmux`.
- **`scripts/dev-static.mjs`**, **`scripts/dev-with-automation.mjs`**: Removed stale `stateDir/tmux` from mkdir lists (no longer needed since tmux dir is under `/tmp`).
- **Tests and docs**: Updated accordingly.

---

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/32eaca87-1a6e-42af-a348-52d3297ff92f)

### DIFF
--- a/scripts/dev-docker.mjs
+++ b/scripts/dev-docker.mjs
@@ -229,7 +229,6 @@ function startAgentServerDocker(config) {
       "/home/openhands/.openhands/agent-canvas/conversations",
     OH_PERSISTENCE_DIR: "/home/openhands/.openhands",
     OH_BASH_EVENTS_DIR: "/home/openhands/.openhands/agent-canvas/bash_events",
-    TMUX_TMPDIR: "/tmp/openhands-agent-canvas-tmux",
     OH_SECRET_KEY: process.env.OH_SECRET_KEY || DEFAULT_SECRET_KEY,
     // Required so the secret-seeding PUT /api/settings/secrets call from
     // the host can authenticate against the agent-server in the container.
--- a/scripts/dev-docker.mjs
+++ b/scripts/dev-docker.mjs
@@ -229,7 +229,7 @@ function startAgentServerDocker(config) {
       "/home/openhands/.openhands/agent-canvas/conversations",
     OH_PERSISTENCE_DIR: "/home/openhands/.openhands",
     OH_BASH_EVENTS_DIR: "/home/openhands/.openhands/agent-canvas/bash_events",
-    TMUX_TMPDIR: "/home/openhands/.openhands/agent-canvas/tmux",
+    TMUX_TMPDIR: "/tmp/openhands-agent-canvas-tmux",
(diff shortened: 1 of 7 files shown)


scripts/dev-docker.mjs
```diff
@@ -229,7 +229,7 @@ function startAgentServerDocker(config) {
       "/home/openhands/.openhands/agent-canvas/conversations",
     OH_PERSISTENCE_DIR: "/home/openhands/.openhands",
     OH_BASH_EVENTS_DIR: "/home/openhands/.openhands/agent-canvas/bash_events",
-    TMUX_TMPDIR: "/home/openhands/.openhands/agent-canvas/tmux",
+    TMUX_TMPDIR: "/tmp/openhands-agent-canvas-tmux",
```
rbren: should we just not set this at all here and leave it default?

xingyaoww: @OpenHands please do it then

xingyaoww: acually i think we can unset bash_events too, so they are saved by default

xingyaoww: Done — removed the explicit `TMUX_TMPDIR` from `containerEnv` in a1589ab. The container's default (`/tmp`) already supports Unix sockets, so no need to set it.

_This reply was generated by an AI agent (OpenHands)._

PR #349, opened 2026-05-11
fix(files-tab): default to Diff view when a workspace or repo is attached

rbren:
## What

When you start a conversation with a workspace or repo selected on the home page, the Files tab now opens in **Diff** mode. Conversations started without an attachment still open in **Files** mode (unchanged).

Previously the Diff default was gated on `selected_repository` only, so conversations created via the workspace picker — which sets a working-dir override but no `selected_repository` — incorrectly landed on Files mode even though there was a pre-existing working tree the user came in to inspect.

## Why this took a few tries

Three previous passes at this logic (see #284 / #288 history) each broke a different case:

| Approach | Symptom |
|---|---|
| Probe via `git status` on the workspace | Every fresh conversation defaulted to Diff, because the agent-server pre-initialises every workspace as a git worktree for its own change tracking. |
| Switch to `selected_repository` only | Fresh no-attachment conversations got the right Files default, but **workspace**-attached conversations regressed to Files because workspace selection isn't recorded as a repo. |
| Add `useHasGitCommits` probe on top of repo-only signal | Fixed empty-repo case but workspace bug still present. |

This PR settles the design as: treat **workspace selection** as semantically equivalent to **repo selection** for the purposes of the Files-tab default, and keep `useHasGitCommits` as the secondary gate so non-git workspaces / unborn-HEAD repos still fall back to Files view.

## How

- **`ConversationMetadata` (`src/api/conversation-metadata-store.ts`)** gains an optional `selected_workspace` field. The agent-server runtime has no concept of workspace selection, so this stays client-side, mirroring how repo metadata already works.
- **Conversation create (`agent-server-conversation-service.api.ts`)** persists conversation metadata when *either* a repo *or* a `workingDirOverride` is supplied (previously: repo only).
- **`useIsGitRepo` → `useHasAttachedSource`** (`src/hooks/use-has-attached-source.ts`): renamed and broadened — returns true when `selected_repository` is on the active conversation **or** `selected_workspace` is in stored metadata.
- **Files tab** keys its Diff default off `hasAttachedSource` instead of `isGitRepo`. The existing `useHasGitCommits` probe still suppresses Diff for non-git or unborn-HEAD trees.
- **Worktree creation** (`worktree: !!metadata?.selected_repository` in `createConversation`) is intentionally *not* extended to workspace selection — for a plain non-git workspace, `git worktree add HEAD` on the server fails and the conversation never starts. We keep that conservative.
- Added an AGENTS.md note documenting the design + warning future agents off the filesystem-probe approach.

## Behaviour matrix after this change

| Conversation started with… | Has commits? | Files-tab landing view |
|---|---|---|
| Repo selected | yes (the usual case) | **Diff** |
| Repo selected | no (empty repo / unborn HEAD) | Files |
| Workspace selected, workspace is a git repo with commits | yes | **Diff** *(new — was Files)* |
| Workspace selected, workspace is a git repo, unborn HEAD | no | Files |
| Workspace selected, workspace is not a git repo | no | Files |
| No attachment | n/a | Files |

User's per-conversation explicit toggle choice always wins, as before.

## Tests

New: `__tests__/hooks/use-has-attached-source.test.ts` covers the four branches (no conversation / repo / workspace / both).

Updated:
- `__tests__/routes/files-tab.test.tsx` — mock renamed and test descriptions reworded to match the broadened semantics.
- `__tests__/api/use-create-conversation-metadata.test.ts` — added an explicit case for the workspace-only persistence path.

```
✓ __tests__/api/use-create-conversation-metadata.test.ts (3 tests)
✓ __tests__/api/conversation-metadata-store.test.ts (3 tests)
✓ __tests__/routes/files-tab.test.tsx (15 tests)
✓ __tests__/hooks/use-has-attached-source.test.ts (4 tests)
Test Files  4 passed (4)   Tests  25 passed (25)
```

`npm run typecheck` clean. ESLint clean on touched files (one pre-existing literal-string warning in a `vi.mock` line is unchanged).

---


rbren:
Heads-up for the review bot: the previous run failed with `'utf-8' codec can't decode byte 0x97 in position 45991: invalid start byte`. That was my fault — when I appended the AGENTS.md note via a shell heredoc the em dash got transcoded to a single Windows-1252 byte (`0x97`) instead of the proper UTF-8 sequence (`0xE2 0x80 0x94`). Force-pushed an amend that swaps it for a real UTF-8 em dash; AGENTS.md now `.decode("utf-8")`s cleanly. Re-running `/codereview` should work.

_Comment posted by an AI agent (OpenHands) on behalf of the user._

PR #424, opened 2026-05-13
fix(frontend): stale backend health errors after recovery

DevinVinson:

---

## Why

Backend health state could get stuck in localStorage after repeated probe failures, leaving a backend red/disabled even after the server recovered. Opening Manage Backends should give the UI a chance to re-check a previously failed backend and clear that stale error state.

## Summary

- add an opt-in one-shot re-probe path for disabled backends in `useBackendsHealth`
- trigger that recovery check when the user opens Manage Backends so recovered servers clear stale persisted health errors
- cap persisted consecutive failure counts and add regression coverage for the hook, store, and modal behavior

## Issue Number

N/A

## How to Test

1. Run `npm run make-i18n`.
2. Run `npx vitest run __tests__/hooks/query/use-backends-health.test.tsx __tests__/api/backend-registry/health-store.test.ts __tests__/components/backends/manage-backends-modal.test.tsx`.
3. Start a new backend only with `dev:extra-backend`. Add it to your agent-canvas and see it with a green dot. 4. End the server and wait till the status turns red.
5. Restart the server and ensure it is running on the same port.
6. Open the Manage Backends modal and see the dot turn from red to green. 

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Focused tests passed locally.

It could make sense to also check for a healthy backend connection when a user switches to that backend but I wasn't sure where best to slot that it.


DevinVinson:
@hieptl I noticed my connected backends were stuck red unless I edited the server itself. I didn't think of the case when writing the bug ticket where the server information was actually correct and the backend was just not running. If this doesn't fit with what you were trying to do by not using `useBackendsHealth` feel free to close this PR and we can fix it a different way.

hieptl (approved):
Thank you! 🙏

PR #432, opened 2026-05-14
Use same-origin backend routing for dev launches

xingyaoww:
## Summary

- Keep browser-side local agent-server calls on the current web origin for dev-launched stacks, so the same stack works through localhost, LAN addresses, and Tailscale hostnames.
- Remove runtime localhost rewriting for explicitly configured backend URLs; configured `localhost` / `127.0.0.1` values now stay direct, which preserves Vercel-preview-to-local-agent-server workflows.
- Apply same-origin dev launch behavior by clearing `VITE_BACKEND_BASE_URL` while keeping `VITE_BACKEND_HOST` as the server-side Vite proxy target.
- Add regression coverage for direct localhost preservation, same-origin fallback, and both dev launcher env models; restore the chat interface snapshot target, refresh CI-generated snapshot baselines, and merge the latest `main` to resolve conflicts.
- Document the dev-stack exposure model and the expectation that externally reachable dev ports should only be exposed on trusted networks.

## Root Cause

The previous URL approach treated browser-facing backend URLs and server-side proxy targets as the same value. Baking `http://127.0.0.1:<port>` into the frontend works only when the browser runs on the same machine as the dev stack. Remote browsers on LAN or Tailscale need to call the current web origin and let the dev server or ingress proxy forward requests server-side.

## Validation

- `npx vitest run __tests__/scripts/dev-safe.test.ts __tests__/scripts/dev-with-automation.test.ts __tests__/api/agent-server-config.test.ts`
- `npm test`
- GitHub checks passed before the scope-only revert; they will rerun on the updated head.


neubig (changes requested):
I'm not sure this is desired behavior. For instance, what if we are viewing a vercel preview and pointing at a local agent server? It seems like this change would prevent that.

xingyaoww:
For some reason 😅 without this PR main just works for me, so gonna close this for now.

PR #439, opened 2026-05-14
Check Docker dev home permissions before launch

xingyaoww:
## Summary

- Add a Docker preflight probe for the mounted `~/.openhands` persistence directory before `npm run dev` starts services.
- The probe uses the same agent-server image and mount target as the real dev container, then verifies the container user can write there and adjust permissions.
- Fail fast with the image user, the failing probe detail, and host-side permission fix commands instead of surfacing a later `PATCH /api/settings` 500.
- Add regression tests for the generated probe command and document the startup check.

## Validation

- `npx vitest run __tests__/scripts/dev-docker.test.ts`
- `PROJECT_PATH=/home/xingyaow/Projects timeout 60s node --env-file-if-exists=.env scripts/dev-docker.mjs` (verified fail-fast permission message before services start)
- `npm exec prettier -- --check scripts/dev-docker.mjs __tests__/scripts/dev-docker.test.ts`


xingyaoww:
Closing this in favor of a smaller follow-up that fixes the Docker launch user model directly.

PR #448, opened 2026-05-14
Auto-refresh stale static frontend

xingyaoww:
## Summary
- detect when an already-open static frontend is still running an older set of `/assets/...` files
- fetch the current page HTML with `cache: no-store` and reload when the served asset fingerprints have changed
- run the check on startup, on focus/visibility changes, and at a small interval
- add focused unit coverage for reload/no-reload behavior and cache-busted HTML checks

## Root Cause
In static dev mode, rebuilding the frontend changes the content-hashed asset URLs served by `build/`, but an already-open browser tab can keep executing the old JavaScript until the user manually refreshes. That can leave the UI using stale baked frontend configuration after `npm run dev` is relaunched.

## Validation
- `npx vitest run __tests__/utils/static-build-refresh.test.ts`
- `npm exec prettier -- --check src/entry.client.tsx src/utils/static-build-refresh.ts __tests__/utils/static-build-refresh.test.ts`

### DIFF
--- a/__tests__/utils/static-build-refresh.test.ts
+++ b/__tests__/utils/static-build-refresh.test.ts
@@ -1,0 +1,103 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  STATIC_BUILD_REFRESH_QUERY_PARAM,
+  refreshIfStaticAssetsChanged,
+} from "#/utils/static-build-refresh";
+
+function htmlResponse(html: string): Pick<Response, "ok" | "text"> {
+  return {
+    ok: true,
+    text: () => Promise.resolve(html),
+  };
+}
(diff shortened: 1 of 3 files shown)


xingyaoww:
Closing this PR because the simpler path is to launch the dev stack in dynamic mode instead of adding frontend stale-build detection for static mode. For local Docker dev, use:\n\n```bash\nnpm run dev:docker:dynamic\n```\n\nThis keeps the frontend on Vite's dev server, so normal reload/HMR behavior handles frontend updates without extra runtime polling.

xingyaoww:
Closing in favor of using the existing dynamic dev launch path.

PR #490, opened 2026-05-15
fix: prevent snapshot CI runs from sending PostHog analytics events

malhotra5:
## Problem

The Playwright snapshot test suite (`npm run test:e2e:snapshots`) was silently leaking `canvas_install` PostHog events on every CI run, inflating the "unique persons" count and making install metrics unreliable.

**Root causes — three things were missing simultaneously:**

1. **Snapshot test `setupMocks` helpers did not set the telemetry localStorage guards** — so on every fresh Playwright browser context `useTelemetry()` would see `openhands-telemetry-first-use` unset, call `trackInstall()`, and fire a `canvas_install` event with a brand-new PostHog `distinct_id`. Since each test file gets its own browser context (= fresh localStorage = new UUID), this was **~27+ `canvas_install` events per CI run** × 3 browser projects (Chromium/Firefox/WebKit), each counted as a new "person" by PostHog.

2. **`ANALYTICS_HANDLERS` (MSW) only intercepted `https://us.i.posthog.com/e`** — the library telemetry service routes through `https://z.openhands.dev` (the OpenHands reverse proxy to PostHog), which MSW never saw.

3. **The PostHog key was always compiled into every build** — no distinction between local dev, CI, and real deployments, so any production-mode build could initialise PostHog.

For comparison, the live E2E tests already had the first two layers of protection (`enableLiveE2EFlags` setting both localStorage keys, and `guardAgainstPostHogRequests` blocking network requests). Snapshot tests had neither.

### 1. All 9 snapshot test `setupMocks` / `addInitScript` helpers
```ts
window.localStorage.setItem("openhands-telemetry-first-use", "true");
window.localStorage.setItem("openhands-telemetry-consent", "denied");
```
Added to every snapshot test file that calls `page.addInitScript()`. Matches the pattern already used in `docker-workspace-browser.snapshot.spec.ts` and `enableLiveE2EFlags` in the live E2E helper. The `onboarding` test is handled correctly — it still removes `openhands-onboarded` to show the onboarding modal, but now also opts out of telemetry.

### 2. `src/mocks/analytics-handlers.ts` — intercept `z.openhands.dev`
```ts
http.post("https://z.openhands.dev/*", async () =>
  HttpResponse.json(null, { status: 200 }),
),
```
Belt-and-suspenders: if PostHog SDK somehow initialises in a mock-mode context, MSW will now swallow the request before it reaches the reverse proxy.

### 3. `src/services/telemetry.ts` — harden PostHog key selection

The PostHog key is **hardcoded per deployment environment** and baked into the static bundle at build time. The staging key is the safe default; the prod key is only used when `VITE_APP_ENV=production` is explicitly set at bundle time.

```ts
const POSTHOG_PROD_KEY    = "phc_BgzfxKdgsYMLFTmJqt424ZoyVHvKFfrwttLimzdYTKFK";
const POSTHOG_STAGING_KEY = POSTHOG_PROD_KEY; // swap for dedicated key when ready

const POSTHOG_API_KEY: string =
  (import.meta.env.VITE_POSTHOG_API_KEY as string | undefined) ||
  (import.meta.env.VITE_APP_ENV === "production"
    ? POSTHOG_PROD_KEY
    : POSTHOG_STAGING_KEY);
```

| Context | `VITE_APP_ENV` | Key compiled in |
|---|---|---|
| `npm run dev` / local builds / CI | unset | `POSTHOG_STAGING_KEY` |
| `npm run build:lib` (npm release) | `"production"` (hardcoded in script) | `POSTHOG_PROD_KEY` |
| Production deployment | `"production"` (set in CI/Vercel) | `POSTHOG_PROD_KEY` |
| Staging deployment | unset or `"staging"` | `POSTHOG_STAGING_KEY` |

`import.meta.env.PROD` was intentionally avoided — it is `true` for any `vite build` output including the static bundle `npm run dev` produces for local use, so it cannot reliably distinguish a real deployment from a local developer build.

`VITE_APP_ENV=production` is baked into `build:lib` directly so every npm package release always uses the prod key without needing the CI pipeline to set it.

## Files changed

| File | Change |
|---|---|
| `src/mocks/analytics-handlers.ts` | Add `z.openhands.dev` MSW intercept |
| `src/services/telemetry.ts` | Staging key as default; prod key only when `VITE_APP_ENV=production` |
| `package.json` | Add `VITE_APP_ENV=production` to `build:lib` |
| `.env.sample` | Document `VITE_APP_ENV` |
| `tests/e2e/snapshots/*.snapshot.spec.ts` (9 files) | Add telemetry localStorage guards to every `setupMocks`/`addInitScript` |


src/services/telemetry.ts
```diff
@@ -32,12 +32,20 @@ const TELEMETRY_CONSENT_KEY = "openhands-telemetry-consent";
...
-// Note: The default API key sends telemetry to OpenHands' PostHog project.
-// Library consumers can override this with their own PostHog project key.
-const POSTHOG_API_KEY =
-  import.meta.env.VITE_POSTHOG_API_KEY ||
-  "phc_BgzfxKdgsYMLFTmJqt424ZoyVHvKFfrwttLimzdYTKFK";
+// PostHog project keys — one per deployment environment, hardcoded so they
+// are baked into the static bundle at build time and cannot drift at runtime.
+// Replace POSTHOG_STAGING_KEY with a dedicated project key once provisioned.
+const POSTHOG_PROD_KEY = "phc_BgzfxKdgsYMLFTmJqt424ZoyVHvKFfrwttLimzdYTKFK";
+const POSTHOG_STAGING_KEY = POSTHOG_PROD_KEY;
```
malhotra5:
```suggestion
const POSTHOG_STAGING_KEY = "phc_kBtz5nKmxVRRQ7HtPwr2QX9eMC5j65zE86QKocVNwb4U";
```

Issue #211, opened 2026-05-09
Public skills are loading inside conversations


malhotra5:
## Description

Public skills are loading inside conversations, which may be unexpected behavior.

## Expected Behavior
Skills should be loaded based on explicit configuration or user intent, not automatically from public sources.

## Current Behavior
Public skills are being loaded within conversations without explicit user action.

## Impact
- Potential unexpected behavior in conversations
- Users may not be aware of which skills are active
- Security/privacy considerations around automatic skill loading

## Steps to Reproduce
1. Start a new conversation
2. Observe that public skills are loaded automatically

## Additional Context
This issue was identified and needs investigation to understand the skill loading mechanism and determine the appropriate fix.

---
*This issue was created by an AI agent (OpenHands) on behalf of the user.*

PR #362, opened 2026-05-12
fix(skills): populate slash menu and skills modal

VascoSch92:
- [x] A human has tested these changes.

## Summary

Fixes #307. The slash menu and skills modal were empty on cloud backends because they fetched skills through a path (`useConversationSkills` → `AgentServerConversationService.getSkills` →
  `loadSkillsForConversation`) that only ever spoke to the local agent-server. On cloud, `createSkillsClient` falls back to the first registered local backend — which usually doesn't exist, so the call returned
  nothing.
  
The settings page already had the right pattern: `SkillsService.getSkills` branches on `getActiveBackend().backend.kind === "cloud"` and delegates to `fetchCloudSkills` via `callCloudProxy`. Agent-canvas has no
  per-conversation skill state (no sandbox-status gating, no org context, no per-conv `disabled_skills` server-side), so the conversation-scoped path was never doing anything the settings path didn't already do
  correctly. This PR consolidates both consumers onto `useSkills`.
  
As a side effect, public skills now appear in the slash menu by default (the old conversation path gated on `VITE_LOAD_PUBLIC_SKILLS`; the settings-page path forces `load_public: true`), matching the second
  bullet of the issue's expected behavior.

## Issue Number
Closes #307

## Tests

- `src/hooks/chat/use-slash-command.ts` and `src/components/features/conversation-panel/skills-modal.tsx`: switched from `useConversationSkills` to `useSkills`.
  - `src/components/features/chat/components/slash-command-menu.tsx`: description extraction falls back to `skill.description` when `content` is absent (cloud `SkillInfo` carries `description` but no `content`);
  narrows via `in` to match `type-guards.ts` style.
  - `src/components/features/conversation-panel/skill-item.tsx`: typed against `SkillInfo`, tolerates missing content/triggers.
  - Dead code removed: `loadSkillsForConversation` (`agent-server-adapter.ts`), `AgentServerConversationService.getSkills`, and the alias hook `use-conversation-skills.ts`.
  - Tests updated to mock `SkillsService.getSkills` (the real data source).

## Video/Screenshots

https://github.com/user-attachments/assets/64398269-6a69-4085-a4d5-40fdfb08e0b9

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

PR #531, opened 2026-05-16
fix: gate public skills in conversations

enyst:
## Summary
- gate `agent_context.load_public_skills` behind `shouldLoadPublicSkills()` when starting conversations
- keep public skills enabled by default so existing conversations retain current behavior
- let deployers opt out with `VITE_LOAD_PUBLIC_SKILLS=false`, while the global Skills page still loads the public catalog
- add regression coverage for the default-on and explicit-opt-out cases

Important: there is no eval risk because default behavior does not change, this only allows users to opt out of public skills if they want to.

## Testing
- npm run make-i18n
- npx --yes vitest run __tests__/api/agent-server-config.test.ts __tests__/api/agent-server-adapter.test.ts __tests__/api/skills-service.test.ts
- npm run typecheck

Refs #211

_This PR was updated by an AI agent (OpenHands) on behalf of the user._

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@
   - `VITE_WORKING_DIR` for the default workspace path sent when starting conversations.
   - `VITE_WORKER_URLS` as a comma-separated list of browser worker URLs if you want the Browser tab to probe exposed app hosts.
   - `VITE_ENABLE_BROWSER_TOOLS=false` to omit `BrowserToolSet` from new conversation payloads.
-  - `VITE_LOAD_PUBLIC_SKILLS=false` to disable loading public skills from the OpenHands extensions marketplace (https://github.com/OpenHands/extensions). Defaults to true.
+  - `VITE_LOAD_PUBLIC_SKILLS=true` to opt conversations into loading public skills from the OpenHands extensions marketplace (https://github.com/OpenHands/extensions). Defaults to false; the global Skills page still forces `load_public=true` so users can browse the public catalog even when conversations keep it disabled.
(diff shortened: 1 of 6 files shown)


smolpaws:
Nice cleanup — gating behind `shouldLoadPublicSkills()` instead of hardcoding is the right move.

One thought: should we default `VITE_LOAD_PUBLIC_SKILLS` to `true` rather than `false`? Right now this PR flips the default from "always on" to "off unless explicitly opted in", which means users who upgrade will silently lose public skills in conversations without knowing why. Defaulting to `true` keeps current behavior while still giving users the knob to turn it off.

smolpaws:
After digging through the #proj-agent-canvas Slack discussion (May 11–15), I think the default here should be flipped to `true`:

1. **Everyone wants skills loaded** — multiple team members actively worked to *fix* skills not loading in conversations. PR OpenHands/agent-canvas#362 was merged to make skills appear in the slash menu and conversation context.

2. **The real problem was performance, not whether to load** — public skills were fetching from the extensions repo on every conversation start, which was slow. Hiep fixed this with SDK-side caching (software-agent-sdk#3253, merged).

3. **The open question is curation, not disabling** — Rohit M asked (May 15): "do we want to cache all openhands extensions skills, or curate a few default skills?" The direction is toward a curated default set, not turning them off.

4. **Nobody in the discussion said "don't load public skills by default"** — the consensus was: load them, make it fast, eventually curate which ones.

Defaulting to `false` would silently remove skills from conversations for all users who upgrade — the opposite of what the team has been working toward. The gating behind `shouldLoadPublicSkills()` is a good change (making it configurable), but the default should stay `true` to match current behavior and team intent.

See also my comment on OpenHands/agent-canvas#2319 — the issue as originally filed may be outdated.

enyst:
Addressed the feedback in `3be30b4`: kept the new `shouldLoadPublicSkills()` gate, but flipped the default back to enabled so existing conversations retain current behavior and `VITE_LOAD_PUBLIC_SKILLS=false` remains the explicit opt-out.

PR #532, opened 2026-05-16
fix: prefer llm profile name in chat footer

enyst:
## Summary
- add a shared footer LLM display resolver that prefers a profile name only when the frontend can resolve it reliably
- show the profile name in the bottom chat LLM chip and overflow submenu, while still surfacing the raw model when resolution is ambiguous
- keep duplicate-profile cases safe by falling back to the model unless a `/model <profile>` switch or other exact signal identifies the profile

## Why
The conversation payload currently exposes `llm_model`, not an explicit profile name. The UI already knows the saved profiles list, the local `active_profile`, and in-session `/model <profile>` switches, so it can often display the friendlier profile name without guessing.

## Testing
- `npm run make-i18n`
- `npx --yes vitest run __tests__/components/features/chat/components/chat-input-model.test.tsx __tests__/components/features/chat/components/chat-input-actions.test.tsx __tests__/components/features/chat/components/use-chat-input-llm-display.test.ts`
- `npm run typecheck`


AGENTS.md
```diff
@@ -430,4 +430,6 @@ return new ConversationClient(getAgentServerClientOptions()).someMethod(...);
...
 
+- Conversation footer LLM chip (`src/components/features/chat/components/chat-input-model.tsx`) should prefer a saved LLM profile name over the raw model only when the frontend can resolve it reliably. The shared resolver lives in `use-chat-input-llm-display.ts`: exact `/model <profile>` switches recorded in `useModelStore` win, the home page may use `active_profile`, otherwise only a unique profile→model match is safe; duplicate profiles that share the same model must fall back to the raw model string.
```
enyst:
```suggestion
- Conversation footer LLM chip (`src/components/features/chat/components/chat-input-model.tsx`) should prefer a saved LLM profile name over the raw model when the frontend can resolve it reliably. The shared resolver lives in `use-chat-input-llm-display.ts`: exact `/model <profile>` switches recorded in `useModelStore` win, the home page may use `active_profile`, otherwise only a unique profile→model match is safe; duplicate profiles that share the same model must fall back to the raw model string.
```

enyst:
@VascoSch92 Does this PR help or makes it more annoying for you? I think maybe we can display profile in a lot of common cases for now, even if we clean it up when your solution lands, but I could be wrong.

enyst:
@openhands try to merge main into this pr; see the very recent commit in main that caused the conflicts and understand it, so you don't miss things. merge main and push.

PR #607, opened 2026-05-18
Fix stale default-local session key after dev:docker restart

### LINKED ISSUE #498: Add backend: 'Login with OpenHands' opens about:blank and /api/cloud-proxy returns 401
### Summary
When adding a backend for **app.all-hands.dev**, clicking **“Login with OpenHands”** opens a new window/tab that stays on `about:blank`. The login flow fails with a **401 Unauthorized** from the local cloud proxy endpoint:

- `http://127.0.0.1:8000/api/cloud-proxy` → **401 (Unauthorized)**

This blocks connecting a cloud backend via the UI.

### Steps to reproduce
1. Run/open the app at `http://127.0.0.1:8000`.
2. Go to **Add backend**.
3. Choose/enter **app.all-hands.dev** as the backend.
4. Click **🔑 Login with OpenHands** (button has `data-testid="add-backend-login-button"`).

### Actual behavior
- Popup/new tab opens and remains on `about:blank` (no redirect to OpenHands auth page).
- Browser console shows:
  ```
  Failed to load resource: the server responded with a status of 401 (Unauthorized)
  127.0.0.1:8000/api/cloud-proxy:1 Failed to load resource: the server responded with a status of 401 (Unauthorized)
  ```

### Expected behavior
- Clicking **Login with OpenHands** should open the OpenHands auth flow and complete successfully.
- `/api/cloud-proxy` should not return 401 during this flow (or, if cloud backends are not supported in this deployment mode, the UI should not present this login option and should show a clear “not supported” message).

### Additional context
Button HTML (for reference):
```html
<button data-testid="add-backend-login-button" type="button" class="p-2 text-sm rounded-sm disabled:opacity-30 disabled:cursor-not-allowed cursor-pointer bg-primary text-[var(--oh-color-base)] hover:opacity-80 w-full">🔑 Login with OpenHands</button>
```

### Environment
- App URL: `http://127.0.0.1:8000`
- Browser/OS: (please fill in)
- How the app is run: (e.g. `npm run dev`, `npm run dev:docker`, etc.)

---
_This issue was created by an AI agent (OpenHands) on behalf of a user._

### LINKED ISSUE #602: BUG: Stale default-local session key causes repeated 401s in runtime probes after dev-docker restart
When running:

```powershell
node --env-file-if-exists=.env .\scripts\dev-docker.mjs
```

agent conversations can start successfully, but background runtime probes such as workspace file listing / local git info repeatedly hit `401` from:

```text
/api/bash/execute_bash_command
```

The browser stack points through:

```text
AgentServerRuntimeService.executeCommand
RemoteWorkspace.executeCommand
HttpClient.post
```

**Why this happens**
`dev-docker` generates/persists a session API key and passes it to both:

- Docker agent-server as `OH_SESSION_API_KEYS_0`
- static frontend as `VITE_SESSION_API_KEY`

Most direct client calls use `getAgentServerClientOptions()`, which correctly prefers the current `VITE_SESSION_API_KEY` for the `default-local` backend.

But `toAppConversation()` sets:

```ts
session_api_key: getEffectiveLocalBackend().apiKey || null
```

If browser localStorage already contains a `default-local` backend with an old non-empty API key, that stale key is copied onto each conversation. Runtime hooks then call:

```ts
AgentServerRuntimeService.executeCommand(
  conversationUrl,
  conversation.session_api_key,
  ...
)
```

That stale per-conversation key overrides the current dev-stack key, so `/api/bash/execute_bash_command` returns 401 on a polling interval.

**What should be fixed**
- Refresh stored `default-local` backend auth when its host still matches the current env-derived default but its API key differs from `VITE_SESSION_API_KEY`.
- Make `toAppConversation()` derive `conversation_url` and `session_api_key` from `getAgentServerClientOptions()` instead of reading `getEffectiveLocalBackend().apiKey` directly.
- Make shared auth header helpers for `default-local` prefer the current configured session key too, so cloud-proxy / LookupSecret paths don’t use stale localStorage auth.

**Acceptance Criteria**
- Restarting `dev-docker` with a rotated/generated session key does not require clearing browser localStorage.
- Runtime polling endpoints such as `/api/bash/execute_bash_command` use the current `VITE_SESSION_API_KEY`.
- Existing custom local backends keep their manually configured API keys unless they are the env-derived `default-local` entry.
- Add regression tests for stale `default-local` localStorage key refresh and `toAppConversation()` session key hydration.

jamiechicago312:
## Summary
Fixes stale `default-local` session API key hydration after `dev:docker` restarts/rotations.

This prevents repeated `401` responses from runtime polling endpoints (e.g. `/api/bash/execute_bash_command`) when the browser still has an older `default-local` API key persisted in localStorage.

Fixes #602.
Fixes #498.

## Changes
- Refresh stored `default-local` backend `apiKey` when its host matches the env-derived default but the key differs.
- `toAppConversation()` now hydrates `session_api_key` (and `conversation_url`) from `getAgentServerClientOptions()` so the configured `VITE_SESSION_API_KEY` wins over stale localStorage.
- Default-local auth headers now prefer the configured session key as well.
- Adds regression tests covering both the storage refresh and conversation hydration.

## How to sync + test locally
```bash
# Clone your fork and check out this branch

git clone https://github.com/jamiechicago312/agent-canvas.git
cd agent-canvas

git fetch origin fix/stale-default-local-session-key-602
git checkout fix/stale-default-local-session-key-602

npm ci
npm test
npm run typecheck

# Run the dev:docker stack
npm run dev:docker
```

### Manual verification for the original bug
1. Start the stack (`npm run dev:docker`) and load the UI once so localStorage gets a `default-local` backend entry.
2. Stop the stack.
3. Rotate the session key (for example by deleting `~/.openhands/agent-canvas/session-api-key.txt` on the host, or by setting a new `SESSION_API_KEY`/`VITE_SESSION_API_KEY`), then restart `npm run dev:docker`.
4. Reload the UI **without clearing browser localStorage**.
5. Confirm background/runtime polling no longer emits repeated `401` errors.

## Test Plan
- `npm test`
- `npm run typecheck`

---
_This PR was created by an AI agent (OpenHands) on behalf of @jamiechicago312._


jamiechicago312:
I fully tested this and it actually solved multiple problems for me. It seems agent canvas is running on my Windows machine excellently with these changes!

src/api/backend-registry/auth.ts
```diff
@@ -7,6 +9,13 @@ import type { Backend } from "./types";
  * token in the `Authorization` header.
  */
 export function buildAuthHeaders(backend: Backend): Record<string, string> {
+  if (backend.kind === "local" && backend.id === DEFAULT_LOCAL_BACKEND_ID) {
+    const configuredSessionApiKey = getAgentServerSessionApiKey();
+    if (configuredSessionApiKey) {
```
enyst, line 14: @openhands Explain this code and PR to me: why isn't api key available in the same way as other backends and what does that have to do with the default local backend id, and with Windows? Understand the problem in the PR description; do you have alternatives?

enyst: Good question. The short version is: `default-local` is *not* a normal persisted backend in the same sense as the others.

For ordinary backends, `backend.apiKey` is the single source of truth, so using it directly is correct. For `default-local`, `backend.apiKey` is really just a cached copy of the current agent-server config (`VITE_SESSION_API_KEY` / `openhands-agent-server-config`) that got seeded into `openhands-backends` at some earlier point. After a `dev:docker` restart, that config can change **outside** the backend registry, while browser localStorage still holds the old `default-local` entry. That is why the key is not always “available in the same way” as other backends: this one has two sources of truth, and the env/config one should win.

That is also why the stable `DEFAULT_LOCAL_BACKEND_ID` matters. It is the marker for “this is the env-derived seeded local backend, safe to rehydrate/special-case”. Without that guard, we would risk overwriting user-created local backends that happen to point at a similar host but intentionally use a different key. So the PR scope is intentionally:
- refresh the stored key only for the env-derived `default-local` entry
- prefer the configured session key for auth/header building on that entry
- hydrate local `conversation.session_api_key` from `getAgentServerClientOptions()` for the same reason

The Windows part is mostly incidental, not architectural. Issue #602 was reported from a Windows + `dev:docker` setup, but the bug is really “browser keeps stale localStorage while the dev stack rotates/regenerates its session key”. That can happen on macOS/Linux too. Windows just happened to be the environment where it was first reproduced and confirmed.

One subtle part of the PR: `getAgentServerClientOptions()` was already preferring the configured session key for `default-local`, but some other paths were not. In particular, `toAppConversation()` copied `getEffectiveLocalBackend().apiKey`, and `buildAuthHeaders()` also fell back to `backend.apiKey`. So after a restart, some calls used the fresh key while runtime polling / cloud-proxy-adjacent paths could still use the stale one. This PR makes those code paths consistent.

Alternatives I considered:
1. **Never persist `apiKey` for `default-local` at all** and always derive it from agent-server-config/env. Conceptually cleaner, but a larger refactor because the backend registry/UI currently treat every backend as a concrete stored object.
2. **Stop storing `session_api_key` on local conversations** and resolve auth at call time from `getAgentServerClientOptions()` / `buildAuthHeaders()`. That would fix the runtime-probe symptom, but not every stale-key path by itself.
3. **Only sync on settings writes** (for example when saving the Agent Server settings page). That misses the `dev:docker` restart case, because the key rotation can happen before the browser writes anything.

So the current approach is basically defense in depth with a small patch surface: heal storage on read, prefer the configured key when building auth, and make local conversation hydration pull from the same source.

_This reply was created by an AI agent (OpenHands) on behalf of @jamiechicago312._

enyst:
> I fully tested this and it actually solved multiple problems for me. It seems agent canvas is running on my Windows machine excellently with these changes!

Ah, nice, thank you! I didn't fully understand the problem so I asked the agent before I saw your comment, but then it's good!

enyst (approved):

PR #631, opened 2026-05-19
fix(files-tab): render HTML previews via srcDoc with allow-scripts

### LINKED ISSUE #626: [Bug] Html is not rendered correctly in the canvas
The agent create an index.html webpage but it seems not to be rendered correctly in the built-in visualizer. If I open the source code in a broswer I can see the correct generated page. 

https://github.com/user-attachments/assets/e0b1671e-730b-45cf-9f99-82be9e191d66

See video attached

VascoSch92:
## Summary
- Closes #626. The HTML preview iframe in `FileContentViewer` was loaded from a `blob:` URL with `sandbox="allow-same-origin"` (no `allow-scripts`), so an agent-generated `index.html` with inline `<style>`/`<script>` or relative-URL assets rendered broken in the canvas while opening the same file in a browser looked correct.
- Switches the HTML branch to `srcDoc` (the decoded file text) with `sandbox="allow-scripts"` so inline CSS/JS and absolute-URL assets behave the same way they do in a browser.

## Security note
Dropping `allow-same-origin` while adding `allow-scripts` is the standard safe-preview posture: scripts can run, but the iframe gets an opaque origin and cannot touch the canvas's cookies, storage, or parent DOM. Net isolation of the parent app is strictly stronger than before (the previous `allow-same-origin` would have been a foot-gun if anyone later added `allow-scripts`).

## Screenshot
Screenshot 2026-05-19 at 16 19 08

## Test plan
- [x] `npx vitest run __tests__/routes/files-tab.test.tsx` — updated assertion now checks `srcdoc` + `sandbox="allow-scripts"`.
- [ ] Manually preview an agent-generated `index.html` with inline `<style>`/`<script>` in the canvas and confirm it matches the browser rendering from the screen recording on #626.

### DIFF
--- a/src/components/features/files-tab/file-content-viewer.tsx
+++ b/src/components/features/files-tab/file-content-viewer.tsx
@@ -130,16 +130,17 @@ export function FileContentViewer({ path, viewMode }: FileContentViewerProps) {
 
   // Text-like content.
   if (mimeType === "text/html" || HTML_LIKE_EXTS.has(getExtension(path))) {
-    // Sandbox the preview iframe. The absence of `allow-scripts` means any
-    // `<script>` (or `onerror=…`, inline event handler, …) inside the
-    // previewed file is inert. This is exactly the safe-preview posture we
-    // want — users can look at their HTML without it executing in the
-    // canvas's context.
+    // Render via `srcDoc` (not a blob: `src`) so inline `<style>`/`<script>`
+    // and absolute-URL assets behave the way they do when the file is opened
+    // directly in a browser. `sandbox="allow-scripts"` lets the page's own
+    // JS run but — crucially — without `allow-same-origin` the iframe gets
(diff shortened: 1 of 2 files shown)


src/components/features/files-tab/file-content-viewer.tsx
```diff
@@ -130,16 +130,17 @@ export function FileContentViewer({ path, viewMode }: FileContentViewerProps) {
...
+    // directly in a browser. `sandbox="allow-scripts"` lets the page's own
+    // JS run but — crucially — without `allow-same-origin` the iframe gets
+    // an opaque origin, so its scripts cannot touch the canvas's cookies,
+    // storage, or parent DOM.
     return (
       <iframe
         title={path}
-        src={staticUrl}
-        sandbox="allow-same-origin"
+        srcDoc={text ?? ""}
```
rbren, line 142: there's a reason this doesn't work well, I have to remember why....

PR #654, opened 2026-05-20
fix: skip git and /projects probes when no workspace is configured

### LINKED ISSUE #642: Bug: 404 and failed bash commands when running first conversation from onboarding with no workspaces setup
Running with a fresh `npm run dev` defaults to no workspaces setup. Each time agent canvas attempts to look for or use a workspace console shows a 404 and the terminal shows a bunch of failed bash commands when running your first conversation from the onboarding prompt.

```
Request URL
http://127.0.0.1:8000/api/file/search_subdirs?path=%2Fprojects
Request Method
GET
Status Code
404 Not Found
Remote Address
127.0.0.1:8000
```
```
22:18:18 [agent-server] [05/19/26 18:18:18] ERROR    Error executing bash command    bash_service.py:335
22:18:18 [agent-server] 'find . -mindepth 2 -maxdepth 4
22:18:18 [agent-server] -name .git 2>/dev/null | sed
22:18:18 [agent-server] 's#^\./##' | sed 's#/.git$##'':
22:18:18 [agent-server] [Errno 2] No such file or
22:18:18 [agent-server] directory: 'workspace/project'
```

nehaprasad-dev:
#### why

- Fresh dev with no workspaces still probed missing workspace/project and /projects, causing 404s and bash errors on first onboarding conversation.

#### Summary

- Gate useLocalGitInfo to attached/partial-repo cases; probe only real paths.
- Query implicit /projects only after a workspace parent is saved.

#### Issue Number
fix : #642

#### How to Test

- npm run dev with no workspaces.
- Finish onboarding “Say hello”.
- No /projects 404 or repeated bash errors in logs.

#### Video/Screenshots
- N/A

#### Type
[x] Bug fix

#### Notes
- Implicit /projects needs at least one saved workspace parent in dev.


DevinVinson (commented):
I think this is close, but I don't think the `useLocalGitInfo` gate is the right predicate yet.

The bug report is specifically about probing a missing default working dir. This PR changes that into "only probe if the conversation has repo metadata or a localStorage-backed attached workspace". That fixes the reported no-workspace spam, but it also drops a previously useful path: conversations that rely on a valid configured working dir (`VITE_WORKING_DIR` / agent-server settings) but do **not** have `selected_repository` or `selected_workspace` metadata. Before this PR, `useLocalGitInfo` would still infer repo/branch from that real working dir; after this PR the query is disabled entirely, so the git control bar loses repo/branch info in that setup.

I’d rather see this gated on whether the candidate working dir is actually probeable / non-default, or have the probe skip failing candidates, instead of keying it to "was a source explicitly attached".

Also worth adding a regression test for the original bash-spam case plus a positive case proving a configured working dir still gets probed.

nehaprasad-dev:
@DevinVinson i preflight dirs with search_subdirs before bash, so configured working dirs still work without metadata, and missing defaults don’t spam bash. Tests cover both cases.

DevinVinson:
@hieptl Can you give this a quick review if you have a chance

smolpaws (commented):
🐾 *smolpaws here — feline code reviewer, OpenHands maintainer team. I read diffs the way cats read laser dots: with intense focus and occasional pouncing.*

---

Hey @nehaprasad-dev — I took a close look at this PR and unfortunately there are some serious structural issues that need addressing before this can move forward.

**Broken function declaration**
The diff removes the `async function probeGitInfo(` line but leaves the parameter list (`run: RunCommand, directory: string`) dangling as bare statements. This is a syntax error — the function body has no function wrapping it.

**Deleted code without replacement**
`GIT_INFO_COMMAND` (the shell script that probes git info) is deleted entirely, but nothing replaces it. The `probeGitInfo` function relied on this — without it, the git probe does nothing.

**Test file has `main` injected**
In `workspace-selection-form.test.tsx`, the string `main` appears on a bare line where code was removed (around line 248). This looks like a merge artifact or an accidental edit.

**Broken `queryFn`**
In `use-local-git-info.ts`, the `queryFn` is left incomplete — it ends with an empty line where the function body should be. The `enabled` property was also removed, which means the query would fire unconditionally.

**Deleted documentation comments**
Several important doc comments explaining *why* the probe exists and how it behaves were removed without replacement. These were useful context for future maintainers.

**`implicitParents` variable deleted but still used**
In `use-resolved-workspaces.ts`, the declaration of `implicitParents` is removed but it's still referenced on the next line (`const extras = implicitParents.filter(...)`) — this is a runtime error.

The underlying idea of gating probes when no workspace is configured (fixing #642) is sound, but the implementation needs a fresh pass. The current diff appears to have deletion artifacts that break compilation. I'd suggest rebasing on the latest `main` and carefully re-applying the changes with the build passing locally before pushing.

CI is failing on both ubuntu and windows builds, which is consistent with these issues.

neubig:

PR #682, opened 2026-05-20
Unify local-backend API key under LOCAL_BACKEND_API_KEY

### LINKED ISSUE #680: Automation API returns 401: frontend sends no API key for /api/automation/* requests
## Problem

The frontend's automation service (`src/api/automation-service/automation-service.api.ts`) reads `VITE_AUTOMATION_API_KEY` for authenticating with the automation backend via `Authorization: Bearer …`. However, the agent-server and automation backend are intended to share the same API key value — only the header format differs (`X-Session-API-Key` vs `Authorization: Bearer …`).

Currently:
- The dev launcher generates **two separate keys** (`sessionApiKey` + `localApiKey`) and persists them to different files
- The Docker entrypoint generates `SESSION_API_KEY` but never sets `OPENHANDS_AUTOMATION_API_KEY`, leaving the automation backend effectively unprotected or inaccessible
- The frontend requires both `VITE_SESSION_API_KEY` and `VITE_AUTOMATION_API_KEY` to be set — if only the session key is configured, automation calls fail with 401

## Expected behavior

Both backends use the same key value. The frontend should only need `VITE_SESSION_API_KEY`.

## Fix

- Frontend: read `VITE_SESSION_API_KEY` for automation Bearer auth instead of `VITE_AUTOMATION_API_KEY`
- Dev launcher: set `localApiKey = sessionApiKey` (no separate generation/persistence)
- Docker entrypoint: default `OPENHANDS_AUTOMATION_API_KEY` / `AUTOMATION_LOCAL_API_KEY` / `AUTOMATION_AGENT_SERVER_API_KEY` to the session key
- Remove `VITE_AUTOMATION_API_KEY` from build env

rbren:
Fixes #680 — same root cause #681 targets, with a slightly broader scope (frontend source-of-truth + one persisted file + a one-shot migration) and a different naming choice. Happy to defer to whichever you prefer; this is the version that landed when I read the issue cold.

## Why

Loading the Automations page produces 401s on `GET /api/automation/v1?limit=50&offset=0` while the rest of the agent-server UI works. Root cause: the dev stack and Docker image maintain two parallel API keys plumbed via two different `VITE_*` vars baked into the frontend:

| Backend       | Server-side env                                    | Frontend bake (VITE)            |
| ------------- | -------------------------------------------------- | ------------------------------- |
| agent-server  | `SESSION_API_KEY` / `OH_SESSION_API_KEYS_0`        | `VITE_SESSION_API_KEY`          |
| automation    | `AUTOMATION_LOCAL_API_KEY`                         | `VITE_AUTOMATION_API_KEY`       |

Each has its own persisted file under `~/.openhands/agent-canvas`, and the two halves silently drift. It's easy to end up in a state where the agent-server accepts the frontend's session key but the automation backend's bearer-token check rejects it (or vice versa) — exactly the 401 you'd see on the Automations page.

## What changes

**One key, end to end.**

```
LOCAL_BACKEND_API_KEY              (server-side env, dev launcher)
  persisted to ~/.openhands/agent-canvas/local-backend-api-key.txt
  (with one-shot migration from the legacy session-api-key.txt)
     -> fed to agent-server as OH_SESSION_API_KEYS_0
     -> fed to automation backend as AUTOMATION_LOCAL_API_KEY
           and AUTOMATION_AGENT_SERVER_API_KEY
     -> baked into the frontend as VITE_LOCAL_BACKEND_API_KEY
          (only used as the npm-run-dev bootstrap)

Frontend source of truth = active backend's stored apiKey
  -> sent as X-Session-API-Key to /api/...                 (agent-server)
  -> sent as Bearer token       to /api/automation/...     (automation)
  -> falls back to VITE_LOCAL_BACKEND_API_KEY only for the
     dev-mode bootstrap before any backend record exists
```

One knob, one persisted file, one frontend secret — the agent-server and automation sidecar accept the same value at the wire, just in different headers.

## Docker (the production case the issue is filed against)

`docker/entrypoint.sh` already generates a `SESSION_API_KEY` and exports `OH_SESSION_API_KEYS_0`, but it never seeds the automation backend's `AUTOMATION_LOCAL_API_KEY`. So the prod image came up with the automation sidecar accepting (effectively) nothing, and the frontend 401'd on every `/api/automation/v1` call regardless of what the user typed into Settings.

This PR defaults `OPENHANDS_AUTOMATION_API_KEY`, `AUTOMATION_LOCAL_API_KEY`, and `AUTOMATION_AGENT_SERVER_API_KEY` to the effective session key. Explicit env overrides still win. (Same fix shape as #681 — credit there.)

## Notes on what did **not** change

- The agent-server still reads `OH_SESSION_API_KEYS_0` and the automation backend still reads `AUTOMATION_LOCAL_API_KEY` / `AUTOMATION_AGENT_SERVER_API_KEY` — those are env var names those backends own.
- The wire-level `X-Session-API-Key` header and `Bearer` Authorization scheme are unchanged.
- Internal TS variables called `sessionApiKey` (e.g. `Backend.apiKey`, `Conversation.session_api_key`) are kept — they refer to the agent-server protocol field, not the env knob.

## Migration / back-compat

- **Dev users** with an existing `~/.openhands/agent-canvas/session-api-key.txt` get a one-shot in-place rename to `local-backend-api-key.txt` on first launch (logged), so their persisted key carries over.
- **Docker users** keep using `SESSION_API_KEY` exactly as before; the entrypoint just now propagates it to the automation env vars too.
- **`OPENHANDS_AUTOMATION_API_KEY` / `AUTOMATION_LOCAL_API_KEY` / `AUTOMATION_AGENT_SERVER_API_KEY`** explicit overrides still work in both dev and Docker.

## Removed input fallbacks (dev launcher only)

`buildSafeDevConfig` / `buildConfig` no longer read these as launcher inputs (they're still set as output env for child processes that need them):

- `SESSION_API_KEY` — use `LOCAL_BACKEND_API_KEY`
- `VITE_SESSION_API_KEY` — use `LOCAL_BACKEND_API_KEY`
- `AUTOMATION_LOCAL_API_KEY` — use `LOCAL_BACKEND_API_KEY`
- `OH_AUTOMATION_API_KEY_PATH` — gone; only one persisted file now

## Test override env vars renamed

| Before                       | After                              |
| ---------------------------- | ---------------------------------- |
| `OH_SESSION_API_KEY_PATH`    | `OH_LOCAL_BACKEND_API_KEY_PATH`    |
| `OH_AUTOMATION_API_KEY_PATH` | _(removed — only one key exists)_  |

## Files touched

20 files, +263 / −289 — dev launcher, Docker entrypoint, frontend automation client (now reads from the active backend), tests, Playwright live config.

All 106 tests in the touched suites (dev-safe, dev-with-automation, dev-static, dev-extra-backend, automation-service, automation-handlers, automations-list, automation-detail, agent-server-config, backend-registry/storage) pass.

## Relation to #681

This PR was developed before I noticed #681. They are largely overlapping:

- Both fix the frontend ↔ automation 401.
- Both fix the Docker entrypoint with the same defaulting pattern.
- Theirs is smaller (+52/-80, 7 files) and back-compat by design — keeps `SESSION_API_KEY` / `VITE_SESSION_API_KEY` and the existing persisted file name. Mine renames the user-facing dev knob to `LOCAL_BACKEND_API_KEY` and adds a migration for the legacy file, plus changes the frontend automation client to read from the active backend's stored apiKey so the Docker case works without any baked-in VITE value.

If the back-compat-first approach is preferred, #681 is the cleaner choice and this can close. If the broader cleanup (one persisted file, no `VITE_AUTOMATION_API_KEY` baking, frontend source-of-truth on the stored backend) is wanted, this is the one — but it merits a closer review of the rename's cost.

---

_This PR was created by an AI agent (OpenHands) on behalf of @rbren._

### DIFF
--- a/__tests__/api/agent-server-config.test.ts
+++ b/__tests__/api/agent-server-config.test.ts
@@ -52,7 +52,7 @@ describe("agent server config", () => {
 
   it("prefills the settings form from environment defaults when local settings are empty", () => {
     vi.stubEnv("VITE_BACKEND_BASE_URL", "https://env-agent.example.com/");
-    vi.stubEnv("VITE_SESSION_API_KEY", "env-session-key");
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "env-session-key");
 
     expect(getAgentServerFormDefaults()).toEqual({
       baseUrl: "https://env-agent.example.com",
@@ -75,7 +75,7 @@ describe("agent server config", () => {
 
   it("lets saved interface settings override environment defaults", () => {
     vi.stubEnv("VITE_BACKEND_BASE_URL", "https://env-agent.example.com");
-    vi.stubEnv("VITE_SESSION_API_KEY", "env-session-key");
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "env-session-key");
 
     saveAgentServerConfig({
       baseUrl: "https://saved-agent.example.com/",
--- a/__tests__/api/backend-registry/storage.test.ts
+++ b/__tests__/api/backend-registry/storage.test.ts
@@ -93,7 +93,7 @@ describe("backend-registry storage", () => {
   });
 
   it("fills a missing API key on the default Local backend from env defaults", () => {
-    vi.stubEnv("VITE_SESSION_API_KEY", "fresh-session-key");
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "fresh-session-key");
     window.localStorage.setItem(
       BACKENDS_STORAGE_KEY,
       JSON.stringify([
@@ -123,7 +123,7 @@ describe("backend-registry storage", () => {
 
 
   it("refreshes a stale API key on the default Local backend from env defaults", () => {
-    vi.stubEnv("VITE_SESSION_API_KEY", "fresh-session-key");
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "fresh-session-key");
     window.localStorage.setItem(
       BACKENDS_STORAGE_KEY,
       JSON.stringify([
@@ -152,7 +152,7 @@ describe("backend-registry storage", () => {
   });
 
   it("does not fill the default Local backend API key after its host is edited", () => {
-    vi.stubEnv("VITE_SESSION_API_KEY", "fresh-session-key");
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "fresh-session-key");
     window.localStorage.setItem(
       BACKENDS_STORAGE_KEY,
       JSON.stringify([
--- a/__tests__/api/to-app-conversation-session-key.test.ts
+++ b/__tests__/api/to-app-conversation-session-key.test.ts
@@ -20,8 +20,8 @@ afterEach(() => {
 });
 
 describe("toAppConversation session_api_key hydration", () => {
-  it("prefers the configured VITE_SESSION_API_KEY over a stale stored default-local apiKey", () => {
-    vi.stubEnv("VITE_SESSION_API_KEY", "fresh-session-key");
+  it("prefers the configured VITE_LOCAL_BACKEND_API_KEY over a stale stored default-local apiKey", () => {
+    vi.stubEnv("VITE_LOCAL_BACKEND_API_KEY", "fresh-session-key");
 
     setRegisteredBackends([
       {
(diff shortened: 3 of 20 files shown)


rbren:
Superseded by OpenHands/agent-canvas#684 — same commit (`005ea2d`), same diff, just renamed the head branch on the fork to drop a misleading `-upstream` suffix. This PR closed automatically when the head branch was renamed (GitHub side-effect). Please review OpenHands/agent-canvas#684 instead.

_This comment was added by an AI agent (OpenHands) on behalf of @rbren._

Issue #640, opened 2026-05-19
Bug: Multiple bug behaviors with LLM profiles


DevinVinson:
This is a bit of a long process to get it to happen but I've been trying to figure out the exact steps. The end result is eventually the API key value gets deleted or corrupted for an LLM profile which causes conversations to start erroring out.

I'm being specific with the models because these are the ones I've definitely seen it happen with. I have a ~5 minute video as well but these are the steps.

1. Add claude-opus-4.5 via OpenHands, check that it works and is your only available profile
2. Add gpt-5.4 via OpenHands
3. See that you have two available profiles. Use claude-opus-4.5 to test conversations work
4. Start a new chat. Switch to gpt-5.4. See when chat starts **the profile is back to claude-opus-4.5** and is not using the profile you selected.
5. Go to LLM Profiles and see that gpt-5.4 is now set as Active.
6. Start a new conversation. See that gpt-5.4 is set as the profile. Send a message and see that claude-opus-4.5 is shown as the profile used.
7. Go to LLM Profiles and **Delete claude-opus-4.5 profile.**
8. Start a new conversation. See that gpt-5.4 is set as the profile. Send a message and see that **the message uses the gpt-5.4 profile successfully** and shows it in the selected profile in chat.
9. Go to LLM Profiles and add a new profile. Select Claude opus 4.6 via OpenHands. See that the API key is *pre-filled* with a green checkmark and <hidden>.
10. Set the new claude-opus-4-6 profile as Active.
11. Start a conversation, see claude-opus-4-6 profile selected. Start the conversation and see the server activity reflect a missing LLM key and no error in the chat session.
12. Go to LLM Profiles. Open the claude-opus-4-6 profile and see there is no API key set.

I think this is 3 specific bugs:
1. The profile shown in the new chat home view isn't always the profile used to start the conversation
1. When adding a new LLM profile the API is sometimes shown with a green checkmark and <hidden> which doesn't seem to actually save an API key
1.  Starting a conversation with an incorrectly configured LLM profile doesn't seem to show an error the user can act on

VascoSch92:
I rechecked this against the current local `agent-canvas` code, and the first two findings look correct.

[...]

PR #711, opened 2026-05-21
fix: surface conversation run errors inline in chat

VascoSch92:

## Why

The agent-server does emit the error — verified by inspecting a failed conversation's persisted events, which contained a ConversationErrorEvent (code: AuthenticationError, full detail). The gap was purely frontend display:
  
- shouldRenderEvent returned false for these event kinds, so they never appeared in the transcript.
- The error banner was only set from the live WebSocket path, never from REST-loaded history — and was fragile anyway (cleared by any subsequent non-error event and on reconnect). Since the run fails fast, by the time the user
  lands on the conversation the error is already in history → nothing shown.

## Summary

Renders ConversationErrorEvent / ServerErrorEvent (e.g. the litellm AuthenticationError raised when an LLM profile has no valid API key) inline in the chat transcript, so a failed run is visible and actionable instead of dying
  silently. 

  Addresses Bug 3 of #640: "Starting a conversation with an incorrectly configured LLM profile doesn't seem to show an error the user can act on."

## Issue Number
#640

## Testing
- New tests for the renderer (icon present, detail expands, ignores non-error events) and shouldRenderEvent (both error kinds render).
- Full conversation-events + contexts suites pass; typecheck, eslint, prettier clean.

## Video/Screenshots

https://github.com/user-attachments/assets/607a9668-0e1b-424f-bddc-b0c40652a331

## Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

| Component | Value |
|---|---|
| **Agent Server** | `ghcr.io/openhands/agent-server:1.23.0-python` |
| **Automation** | `openhands-automation==1.0.0a5` |
| **Commit** | `a8de7cc9d85862f1fd9abba936b8a0b10f859841` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-a8de7cc
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-a8de7cc
```

```
ghcr.io/openhands/agent-canvas:sha-a8de7cc-amd64
ghcr.io/openhands/agent-canvas:vasco-error-msg-amd64
ghcr.io/openhands/agent-canvas:pr-711-amd64
ghcr.io/openhands/agent-canvas:sha-a8de7cc-arm64
ghcr.io/openhands/agent-canvas:vasco-error-msg-arm64
ghcr.io/openhands/agent-canvas:pr-711-arm64
ghcr.io/openhands/agent-canvas:sha-a8de7cc
ghcr.io/openhands/agent-canvas:vasco-error-msg
ghcr.io/openhands/agent-canvas:pr-711
```

- Each tag (e.g., `sha-a8de7cc`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-a8de7cc-amd64`) are also available if needed


src/components/conversation-events/chat/event-message.tsx
```diff
@@ -162,6 +165,11 @@ export function EventMessage({
     return <ErrorEventMessage event={event} {...commonProps} />;
   }
 
+  // Conversation/server error events (e.g. an LLM AuthenticationError)
+  if (isConversationErrorEvent(event) || isServerErrorEvent(event)) {
```
enyst, line 169: Just above this, I see agent error events, those fleeting errors like LLM sending the wrong parameters for a tool call. They’re very different than conversation/server errors, in that the first aren’t even errors from the perspective of the application - the agent will just continue and do presumably better next step.

In the video though, it seems this displays the “real” errors under the name “agent error”, and I wonder if we can avoid doing that?

enyst (changes requested):
Thank you for this PR! This is a headache 😓

Code looks good, I do have a tiny question though: I think the video shows that we tell the user and ourselves that conversation/server errors are “agent errors”, but they are not… Confusing them has been a painful source of bugs. Could we display somehow that conversation errors are conversation errors, while agent errors are agent errors?

I sympathize with giving up the banner… on the other hand, it kinda was fulfilling the purpose to make a distinction somehow?

VascoSch92:
@openhands resolve the conflict and update the branch

PR #716, opened 2026-05-21
fix: Windows Bad Gateway — Vite killed by unquoted path in spawnService shell

### LINKED ISSUE #715: Windows: `npm run dev` Bad Gateway — Vite exits with `C:\Program` path error
## Bug

On Windows, `npm run dev` starts the agent-server, automation backend, and ingress proxy successfully, but the UI at `http://localhost:8000/` returns **Bad Gateway** because Vite exits immediately.

```
[vite] Starting on port 3001...
[vite] 'C:\Program' is not recognized as an internal or external command,
[vite] operable program or batch file.
[vite] Exited with code 1
[ingress] Proxy error for /:
```

The ingress proxy's default route targets `http://localhost:3001`. With Vite dead, requests to `/` fail.

## Root Cause

A path under `C:\Program Files\...` is invoked through a Windows shell without quoting, causing `cmd.exe` to execute `C:\Program` instead of the full path.

## Workarounds

**Bypass the npm wrapper (Vite dev mode):**
```powershell
$env:PYTHONUTF8 = "1"
node --env-file-if-exists=.env .\scripts\dev-with-automation.mjs
```

**Use the static frontend launcher:**
```powershell
$env:PYTHONUTF8 = "1"
node --env-file-if-exists=.env .\scripts\dev-static.mjs
# Skip rebuild after first build:
node --env-file-if-exists=.env .\scripts\dev-static.mjs --skip-build
```

## Fix

In `spawnService()`, avoid spawning frontend commands through `cmd.exe` when the command path may contain spaces — either quote the path or set `shell: false` for Node/npm script invocations on Windows.

jamiechicago312:
Fixes #715

## Root Cause

`buildNpmScriptCommand()` in `scripts/dev-safe.mjs` returns `process.execPath` as the command when `npm_execpath` is set (i.e. when launched via `npm run dev`):

```js
// dev-safe.mjs:752
if (env.npm_execpath) {
  return {
    command: env.npm_node_execpath || nodeExecPath,  // e.g. C:\Program Files\nodejs\node.exe
    args: [env.npm_execpath, "run", scriptName],
  };
}
```

`spawnService()` in `scripts/dev-with-automation.mjs` then spawns that command with `shell: true` on Windows:

```js
// dev-with-automation.mjs:460
shell: process.platform === "win32",
```

Node passes the command to `cmd.exe` unquoted, so `cmd.exe` sees `C:\Program` as the executable and errors out — killing Vite before the ingress starts, which causes Bad Gateway on `localhost:8000`.

This affects **any Windows user whose Node.js is installed in `C:\Program Files\`** (the default installer path), regardless of clean or existing install.

## Changes

- **README**: adds a one-paragraph Windows workaround note with the direct-script fallback (`dev-with-automation.mjs`)

## Suggested code fix (follow-up)

Remove `shell: process.platform === "win32"` from `spawnService()`. When `npm_execpath` is set, `buildNpmScriptCommand` already returns a full `node.exe` path — `shell: false` spawns it correctly. When it isn't set (Windows non-npm path), `buildNpmScriptCommand` returns `cmd.exe` as the command explicitly, so no shell wrapping is needed there either.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ npm install
 npm run dev
 ```
 
+> **Windows:** `npm run dev` may show a Bad Gateway error if Vite exits with a `C:\Program` path error. Run the dev stack directly instead:
+> ```powershell
+> $env:PYTHONUTF8 = "1"
+> node --env-file-if-exists=.env .\scripts\dev-with-automation.mjs
+> ```
+
 Access the UI at [http://localhost:8000](http://localhost:8000). You can add additional backends directly from the UI.
 
 # Architecture


README.md
```diff
@@ -56,6 +56,12 @@ npm install
 npm run dev
 ```
 
+> **Windows:** `npm run dev` may show a Bad Gateway error if Vite exits with a `C:\Program` path error. Run the dev stack directly instead:
+> ```powershell
+> $env:PYTHONUTF8 = "1"
+> node --env-file-if-exists=.env .\scripts\dev-with-automation.mjs
```
malhotra5, line 62: Hmm npm run dev simply wraps `node --env-file-if-exists=.env scripts/dev-with-automation.mjs` command

Unclear to me why one fails but the other works

malhotra5: Is there a workaround that allows npm run dev to work? Because once we release the npm package it would be great if they could use it directly

jamiechicago312: That is the thing. I got this error for the first time I used agent canvas last week. And then earlier this week it worked great. and then i got bad getaway again!

@openhands can you investigate the codebase. What would need to change so windows, mac, and linux can run without running into bad getaway at all so that a user after npm install can just run npm run dev

jamiechicago312:
Superseded by OpenHands/agent-canvas#734, which fixes the root cause in code rather than documenting a workaround.

The actual bug was a one-line ordering issue in `buildNpmScriptCommand`: the `npm_execpath` branch fired before the `win32` branch, so on Windows the function returned the full `C:\Program Files\nodejs\node.exe` path as the spawn command. `spawnService` uses `shell:true` on Windows, and Node.js passes that unquoted path straight to `cmd.exe`, which splits on the space and fails with `‘C:\Program’ is not recognized`.

Closing this in favour of OpenHands/agent-canvas#734.

_This comment was added by an AI agent (OpenHands) on behalf of the user._

PR #720, opened 2026-05-21
feat(skills): redesign available skills modal layout and grouping

FraterCCCLXIII:
## Summary

Polishes the **Available Skills** modal (`data-testid="skills-modal"`) opened from the conversation tools menu, chat actions, and conversation name menu.

### Header
- Adds a top-right **Close (X)** button (`close-skills-modal`)
- **Refresh** is icon-only (18px), always visible, disabled only while loading/refetching; uses the same hover styling as close with tooltip + `aria-label`
- **Warning subline** is always shown (no longer gated on agent ready)
- Title + warning use `gap-2` to match other modals

### List layout & grouping
- Modal body and list container use theme border tokens; list sits on `bg-surface-raised` (one step above modal `bg-base-secondary`)
- Skills are grouped into **Project skills → User skills → Public skills** via new `skill-scope.ts` path heuristics and sticky section headers with count pills
- Mutually exclusive content states: runtime waiting → loading → empty/error → grouped list

### Skill rows
- Flat list with `divide-y` dividers (no inner row padding)
- **Monochrome pills** for skill type and triggers (not colored `SkillTypeBadge` / trigger-chip styles)
- Content block uses `bg-base`, `text-xs`, and theme border tokens; trigger spacing aligned with content (`mt-2`)

### Runtime waiting
- New waiting state shows spinner + “Waiting for runtime…” while agent is `LOADING`/`INIT` (fixes missing spinner/subline/refresh during provisioning)

### Global tweak (note for reviewers)
- `BaseModalTitle` now uses `text-content-2` instead of default foreground — affects all modals that use this shared title component

### Section label terminology

The modal introduces scope-based grouping for the first time in the UI. Labels were chosen to align with existing backend/API naming rather than invent new terms:

| Section | Label | Rationale |
|---|---|---|
| `project` | **Project skills** | Matches `load_project` / workspace-scoped skills |
| `personal` (internal) | **User skills** | Matches `load_user`, `load_user_skills`, `source: "user"` |
| `public` | **Public skills** | Matches `load_public`, `VITE_LOAD_PUBLIC_SKILLS`, `source: "public"` |

**Alternatives considered (not used):**
- *Global skills* — friendlier for `~/.openhands` / `~/.agents` paths, but not API-aligned
- *Default skills* — matches Skills settings page copy (“enable or disable default skills”), but reads as toggle state rather than origin
- *Built-in skills* — collides with other “built-in” UI (MCP transport, ACP providers, slash commands)
- *System skills* — clear hierarchy, but not used in API/docs and overlaps with “system prompt” concepts

Internal code keeps `SkillScope = "project" | "personal" | "public"`; only user-facing headers use User/Public.

## Test plan

- [ ] Open skills modal while runtime is starting → spinner, warning, refresh, and close are all visible
- [ ] After runtime is ready → skills appear in Project / User / Public sections with count pills
- [ ] Expand a skill → monochrome type/trigger pills, darker content block, consistent spacing
- [ ] Refresh refetches skills; close dismisses the modal
- [ ] `npm test -- __tests__/components/modals/skills/skill-modal.test.tsx __tests__/utils/skill-scope.test.ts`

## Follow-ups (out of scope)

- Proper translations for new section i18n keys (currently English duplicated across locales)
- Align `HooksModal` header/spacing with the same patterns
- Consider a backend `scope` field instead of path-based heuristics in `skill-scope.ts`


FraterCCCLXIII:
### Naming note: skill section headers

The Available Skills modal is the first place we group skills by *origin* (not skill type). Early drafts used **Personal** and **Built-in**, but those don't match how the rest of the app talks about skills.

**What we landed on:** Project → User → Public — aligned with `load_project` / `load_user` / `load_public` and `source` values from the agent-server.

**Why not the other options?**
- **Global** — nicer for home-dir paths, but not API language
- **Default** — already on `/skills` for toggling catalog skills, but sounds like on/off state
- **Built-in** — already means MCP/ACP/slash commands elsewhere
- **System** — readable, but not in API/docs and easy to confuse with system prompts

Happy to revisit if we want UI copy to follow the Settings page ("default skills") instead of the API ("public skills").

PR #734, opened 2026-05-22
fix: buildNpmScriptCommand always uses cmd.exe on Windows

jamiechicago312:
## Problem

`npm run dev` fails immediately on Windows with:

```
[vite] 'C:\Program' is not recognized as an internal or external command,
[vite] operable program or batch file.
[vite] Exited with code 1
```

Vite never starts, and the ingress proxy returns **Bad Gateway** for every request to `localhost:8000`. This happens on a clean install — no `.env`, no special config needed.

## Root cause

When `npm run dev` is executed on Windows, npm sets `npm_execpath` to something like:

```
C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js
```

In `buildNpmScriptCommand` the **`npm_execpath` branch fires first**, returning that full path (with spaces) as a spawn argument, and using `process.execPath` (`C:\Program Files\nodejs\node.exe` — also with spaces) as the command.

`spawnService` uses `shell: process.platform === "win32"`, so Node.js builds the shell command:

```
cmd.exe /d /s /c C:\Program Files\nodejs\node.exe C:\Program Files\...
```

`cmd.exe` splits on the space, sees `C:\Program` as the command, and fails. The `win32` branch in `buildNpmScriptCommand` already has the correct fix (`cmd.exe /d /s /c npm run <script>`), but it **came after** the `npm_execpath` branch, so it was never reached on Windows when running via `npm run dev`.

## Fix

Move the `win32` check **before** the `npm_execpath` check. On Windows, we always use:

```js
{ command: env.ComSpec || "cmd.exe", args: ["/d", "/s", "/c", "npm", "run", scriptName] }
```

This is safe on all Windows Node installs regardless of where npm was installed or whether `npm_execpath` contains spaces.

No changes needed on macOS or Linux — those platforms continue to use the `npm_execpath`-aware path.

## Testing

- Added a new regression test that specifically exercises **Windows + `npm_execpath` with spaces** (the exact scenario from the bug report).
- Renamed the previous Windows+`npm_execpath` test to POSIX since the npm_execpath shortcut only matters on non-Windows platforms.
- All existing cmd.exe fallback tests continue to pass.

Closes / supersedes #716 (the docs-only workaround).

---

@jamiechicago312 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/377e79a5-43c1-4ab7-902c-8a7ef7eed067)


jamiechicago312:
Logs to proof this fix is successful and a screenshot

[20260522-pr734-succesful-logs.txt](https://github.com/user-attachments/files/28157093/20260522-pr734-succesful-logs.txt)

image

jamiechicago312:
Just updated branch and tested again and it seems to work

https://github.com/user-attachments/assets/d200e593-8315-469b-8b9c-8ae4b03ee1fd

hieptl (approved):
Thank you! 🙏

PR #973, opened 2026-06-01
fix: surface dynamic_context in Agent Tools & Metadata modal (#932)

### LINKED ISSUE #932: [Bug]: Agent Tools & Metadata modal omits `dynamic_context` from the system prompt
### Situation
The "Agent Tools & Metadata" modal shows the system prompt by rendering `system_prompt.text` from the backend `SystemPromptEvent` (via `adaptSystemMessage` in `src/utils/system-message-adapter.ts`).

### Problem
The backend `SystemPromptEvent` actually carries **three** content fields: `system_prompt`, `tools`, and `dynamic_context`. The frontend only reads the first two — `dynamic_context` is never referenced anywhere (`grep dynamic_context src/` returns zero matches), and it's not even in the `SystemPromptEvent` TypeScript interface (`src/types/agent-server/core/events/system-event.ts`).

`dynamic_context` is part of the system message the model actually receives, and it's larger than the static prompt. In a real persisted event:

- `system_prompt.text` ≈ 14,400 chars (static base prompt — shown ✅)
- `dynamic_context.text` ≈ 17,200 chars (**dropped ❌**)

The dynamic context contains the runtime-injected sections: `<CURRENT_DATETIME>`, `<SKILLS>` (full skills catalog), `<RUNTIME_SERVICES>`, and `<CUSTOM_SECRETS>`.

**Result:** the modal displays less than half of the agent's true system context, so what users see does not match what the agent actually receives.

### Suggested fix
- Add `dynamic_context?: TextContent` to the `SystemPromptEvent` interface.
- Surface it in `adaptSystemMessage` / the modal (appended to the prompt or as its own section).
- Note: `dynamic_context` includes a `<CUSTOM_SECRETS>` block; secret values appear masked (`<secret-hidden>`) in the sample, but confirm that holds before exposing it in the UI.

burakkeless:
Closes #932.

The "Agent Tools & Metadata" modal rendered only `system_prompt.text` and `tools` from the backend `SystemPromptEvent`. That event actually carries a third content field, `dynamic_context`, which was never referenced in the frontend (nor in the `SystemPromptEvent` TS interface).

`dynamic_context` is **part of the same system message the model receives** — the final system prompt is effectively `system_prompt + dynamic_context` — and it's larger than the static prompt (≈17.2k vs ≈14.4k chars in a real event). It holds the runtime-injected sections: `<CURRENT_DATETIME>`, `<SKILLS>`, `<RUNTIME_SERVICES>`, and `<CUSTOM_SECRETS>`. As a result the modal showed less than half of the agent's true system context.

## Changes

- **Type:** add optional `dynamic_context?: TextContent` to the `SystemPromptEvent` interface (optional, so older persisted events stay valid).
- **Adapter:** since `dynamic_context` is the runtime-injected tail of the same prompt, `adaptSystemMessage` now **appends** it to `system_prompt.text` (joined with a blank line; the static prompt is `trimEnd()`-ed so the seam is always exactly one blank line). The System Message tab therefore shows the full prompt as the model actually receives it — no new tab.
- **Secrets:** per the issue's note, added a defensive client-side `redactCustomSecrets` backstop that re-masks any unmasked value inside a `<CUSTOM_SECRETS>` block (`KEY=value` / `KEY: value`, tolerant of a truncated block) before display. The backend is still the primary masker (`<secret-hidden>`); this is a safety net only.

## Testing

- `redact-custom-secrets.test.ts` (new): both separators, multiple secrets, truncated block, idempotency, leaves text outside the block untouched.
- `system-message-adapter.test.ts`: `dynamic_context` absent → `content` unchanged; present → appended to `content`; unmasked secret → redacted in `content`.
- `npm run typecheck` and the targeted vitest files pass locally (13 tests).

## Notes

Intentionally dropped versus an earlier approach:

- **No "Dynamic Context" tab** — splitting one continuous system prompt across two tabs would misrepresent it as two separate artifacts. Appending into the System Message tab is the faithful representation.
- Consequently, **no `DYNAMIC_CONTEXT_TAB` i18n key** and **no modal reset-on-open** (that reset only existed to guard the now-removed tab).


burakkeless:
https://github.com/user-attachments/assets/5633b97a-e0a8-49a6-b504-0611e70805cd

VascoSch92:
Hey @burakkeless,

Thank you very much for the video and the PR!

I think I wrote something wrong in my issue. 

The dynamic context is actually part of the system prompt, meaning the final system prompt is basically `= system_prompt + dynamic_context.
`

So, I believe the best thing to do is to append the dynamic context to the system_prompt inside the system message.

Sorry for the misunderstanding!

burakkeless:
https://github.com/user-attachments/assets/bf4f19b2-b4b0-4d3c-ba79-c0b4d356be0e

VascoSch92 (approved):
LGTM.

If you can fix the two suggestions, it would be great ;-)

PR #977, opened 2026-06-01
fix: set AUTOMATION_AGENT_SERVER_URL in Docker entrypoint to enable local auth

tofarr:
## Problem

When running via Docker, the `/automations` view returns 401 on `GET /api/automation/v1` even though the `x-session-api-key` header is sent with the correct value. The same key works fine with the agent server and with `npm run dev`.

## Root Cause

The automation service has two authentication paths in `authenticate_request()` (`openhands/automation/auth.py`):

1. **Local-mode fast path** — validates the key directly against `AUTOMATION_LOCAL_API_KEY`, no network call. Active when `ServiceSettings.is_local_mode` is `True`.
2. **Cloud validation** — forwards the key to `https://app.all-hands.dev/api/v1/users/me`. Returns **401** for locally-generated session keys.

`is_local_mode` is defined as `bool(self.agent_server_url)`, which reads from `AUTOMATION_AGENT_SERVER_URL`. The Docker entrypoint set `AUTOMATION_LOCAL_API_KEY` (the key to validate against) but never set `AUTOMATION_AGENT_SERVER_URL` (the flag that activates local mode), so `is_local_mode` was always `False` and every request fell through to the cloud path.

`npm run dev` (via `dev-with-automation.mjs`) was setting both variables correctly — Docker was the only place missing `AUTOMATION_AGENT_SERVER_URL`.

## Fix

Export `AUTOMATION_AGENT_SERVER_URL` in `docker/entrypoint.sh`, pointing at the agent server’s internal address (`127.0.0.1:$AGENT_SERVER_PORT`). Both services run inside the same container so loopback is correct (not `host.docker.internal`).

```bash
export AUTOMATION_AGENT_SERVER_URL="${AUTOMATION_AGENT_SERVER_URL:-http://127.0.0.1:${AGENT_SERVER_PORT}}"
```

This makes `is_local_mode` return `True`, activating the local fast path that validates the session key against `AUTOMATION_LOCAL_API_KEY` — which is already set to the same session key.

## Screenshots

image

---

| Component | Value |
|---|---|
| **Automation** | `openhands-automation==1.0.0a5` |
| **Commit** | `be29a93ec759d65f0536056d67533b613077cef6` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-be29a93
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-be29a93
```

```
ghcr.io/openhands/agent-canvas:sha-be29a93-amd64
ghcr.io/openhands/agent-canvas:fix-docker-automation-401-amd64
ghcr.io/openhands/agent-canvas:pr-977-amd64
ghcr.io/openhands/agent-canvas:sha-be29a93-arm64
ghcr.io/openhands/agent-canvas:fix-docker-automation-401-arm64
ghcr.io/openhands/agent-canvas:pr-977-arm64
ghcr.io/openhands/agent-canvas:sha-be29a93
ghcr.io/openhands/agent-canvas:fix-docker-automation-401
ghcr.io/openhands/agent-canvas:pr-977
```

- Each tag (e.g., `sha-be29a93`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-be29a93-amd64`) are also available if needed


malhotra5 (approved):
LGTM

PR #1052, opened 2026-06-02
fix(sidebar): scroll conversation filter menu on short viewports

FraterCCCLXIII:
## Problem

The sidebar **Filter conversations** dropdown (`older-conversations-filter-menu`) renders a tall panel with Organize, Sort by, Show, Metadata, Older conversations, and Delete all sections. It had no height constraint, so on short viewports the menu extended below the visible window. Users could not scroll to reach options at the bottom (e.g. toggles and **Delete all**).

## Fix

- Add a shared `dropdownMenuViewportScrollClassName` utility that caps menu height to `min(60vh, calc(100dvh - 5rem))` so the panel stays within the viewport.
- Apply `overflow-y-auto` with the existing `custom-scrollbar` styling on the filter menu.
- Scrollbar appears **only when content overflows** — if the full menu fits, no scrollbar is shown.

## Test plan

- [ ] Open the app with the sidebar visible and click the filter (list-filter) icon next to **Conversations**.
- [ ] At normal window height, confirm the full menu is visible with no scrollbar.
- [ ] Shrink the window height until the menu would exceed the viewport; confirm a scrollbar appears and all sections remain reachable.
- [ ] Verify keyboard navigation (Arrow Up/Down, Escape) still works while scrolling.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/components/features/conversation-panel/conversation-panel-filter-menu.tsx
+++ b/src/components/features/conversation-panel/conversation-panel-filter-menu.tsx
@@ -24,6 +24,7 @@ import {
   dropdownMenuRowIconClassName,
   dropdownInstantColorClassName,
   dropdownMenuListClassName,
+  dropdownMenuViewportScrollClassName,
 } from "#/utils/dropdown-classes";
 import type {
   ConversationSortField,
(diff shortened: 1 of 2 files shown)


hieptl (approved):
Thank you! 🙏

PR #1092, opened 2026-06-03
fix: patch local automation backend compatibility

jamiechicago312:
## Summary
- inject a local automation compatibility shim into the source-install dev launchers
- patch the pinned automation backend at process start so local tarball uploads use the current API file upload route
- use a real local temp tarball path for localhost agent-server runs on Windows and cover the launcher wiring with tests
- include tzdata in the uvx-launched automation runtime so cron polling works on Windows and minimal Python installs

## Problem
The source-install automation stack currently launches openhands-automation==1.0.0a5, whose upload helper still posts tarballs to the stale query-param upload route. Current local agent-server builds expect the path-based upload route instead, so prompt automations fail before they can start a conversation. On Windows local runs, the stale /tmp/automation.tar.gz destination is also not a valid absolute host path.

Separately, the cron scheduler can crash on Windows and minimal Python installs because the uvx environment does not include tzdata, so zoneinfo cannot resolve UTC and polling automations never run.

## Fix
This PR keeps the released automation package pin in place, but adds local-launcher compatibility guardrails:
- dev-with-automation.mjs and dev-static.mjs prepend scripts/automation-compat-python to PYTHONPATH
- the shim sitecustomize.py monkeypatches openhands.automation.execution at interpreter startup
- patched local runs upload via the current file upload route and use a concrete local temp tarball path on Windows localhost runs
- the shared automation uvx command now installs tzdata with --with tzdata so cron triggers can resolve UTC and other IANA zones reliably

## Validation
- npx vitest run __tests__/scripts/dev-with-automation.test.ts
- uvx --from openhands-automation==1.0.0a5 --with tzdata python -c "from zoneinfo import ZoneInfo; print(ZoneInfo('UTC'))"

## Notes
This is intentionally a launcher-side stopgap until openhands-automation publishes a release compatible with the current local agent-server file-upload route and timezone expectations for local source installs.


neubig:

PR #1113, opened 2026-06-04
docs: sync README Docker tags with config/defaults.json

### LINKED ISSUE #1073: [Bug]: README should reference the latest docker version for download
The README and [Docs](https://docs.openhands.dev/openhands/usage/agent-canvas/setup) should reference the latest docker version we want users to download. Currently both of these doc locations are:
- inconsistent (e.g., README points to a pinned version, whereas the Docs point users to `latest`)
- out of date (README points to an older version)

We should either
- cite a major version
- or have a background job that keeps these in sync

nehaprasad-dev:
#### Why

- README Docker install examples pointed at stale, mismatched tags (1.0.0-beta.6 vs 1.0.0-alpha.10) instead of the current release.

#### Summary

- Added versions.agentCanvas to config/defaults.json as the single source of truth for the recommended Docker tag
- Updated README.md and README.windows.md to ghcr.io/openhands/agent-canvas:1.0.0-rc.1
- Extended docs-version-sync.test.ts and the release skill so future bumps cannot drift silently

#### Issue Number
fix : #1073

- npm test -- __tests__/scripts/docs-version-sync.test.ts
- Confirm README.md and README.windows.md both reference ghcr.io/openhands/agent-canvas:1.0.0-rc.1.

#### Video/Screenshots
- N/A — docs and test-only change.

#### Type
- Bug fix


hieptl (changes requested):
Hello @nehaprasad-dev,

Thank you for creating this pull request.

I’ve identified a few issues, which I’ve outlined below for your review and reference.

## Issues Found

🔴 **1. The pinned version is already stale**
v1.0.0-rc.2 was released on 2026-06-04, after this PR pinned rc.1. The PR should be updated to 1.0.0-rc.2 before merging. Ironically, this highlights the structural gap noted in point 3 below.

🟡 **2. Test allows stale tags to coexist**
The test uses `toContain`, which only verifies that the correct tag appears somewhere. Since `README.windows.md` references the image twice, a future update could modify one reference while missing the other, and the test would still pass. A stronger assertion could be:

```ts
for (const file of ["README.md", "README.windows.md"]) {
  const refs =
    read(file).match(/ghcr.io\/openhands\/agent-canvas:[^\s\`"]+/g) ?? [];
  expect(refs.length).toBeGreaterThan(0);
  for (const ref of refs) expect(ref).toBe(dockerImage);
}
```

🟡 **3. No mechanism keeps `config/defaults.json` in sync with releases**
The current test only guards against drift between the README and config. However, synchronization between `config/defaults.json` and the latest release depends entirely on manual adherence to `release.md`, which has already drifted (see point 1) within a day.

Issue 1073 explicitly proposed automation here. A lightweight safeguard could be adding a step in `create-release.yml` that fails (or opens an automated PR) if `versions.agentCanvas` does not match the tag being released.

🟡 **4. Issue 1073 is only partially resolved**
The issue references two documentation surfaces: the README and the external Docs site (docs.openhands.dev), which points users to “latest.” This PR only addresses the README portion.

Additionally, `docker.yml` only publishes a `latest` tag for stable releases (lines 152–164), while all current releases are pre-releases. As a result, the Docs site’s “latest” reference may already be broken for users. This likely requires a follow-up change in the docs repository or an explicitly tracked follow-up issue. It would be helpful to note this explicitly so the issue is not prematurely closed as fully resolved.

🔵 **5. Minor nits in `release.md` (mostly pre-existing)**

- The `sed -i 's/...:[^ ]*/...'` pattern could inadvertently consume a closing backtick if the image reference appears inside inline code. Additionally, `sed -i` is not compatible with macOS BSD sed without `sed -i ''`. This appears pre-existing, but this PR touches the line.
- Step 1 instructs manually setting `versions.agentCanvas`, but the snippet includes `git add config/defaults.json` even though the sed command does not modify it, which may cause slight confusion in execution order.
- The documentation update commit is applied on the `rel-<X.Y.Z>` branch per the release process. However, GitHub users primarily view the README on `main`. It may be worth confirming that release branches are consistently merged back; otherwise, `main` may continue to serve stale documentation even when the release flow is completed correctly.

Thank you very much! 🙏

hieptl (approved):
Thank you! 🙏

PR #1117, opened 2026-06-04
fix: mask saved LLM API key field with fixed-length placeholder instead of stored-value-length dots

### LINKED ISSUE #1078: API key field displays longer than actual key after saving, making it hard to edit
### Description

Once an API key has been saved, the password field in the LLM settings renders significantly more dots than the actual length of the stored key. The field appears much wider/longer than it should, which makes it visually confusing and hard to select/clear the field when trying to update the key.

### Steps to Reproduce

1. Open the LLM settings panel
2. Enter and save an API key
3. Re-open the settings panel
4. Observe the API key field — the dots stretch well beyond the actual key length

### Expected Behavior

The masked field should reflect the actual length of the stored key (or use a fixed-width placeholder like `••••••••` to indicate a key is set), so users can easily tell when the field is populated and can confidently clear/replace it.

### Actual Behavior

The dots rendered in the field are considerably longer than the real key value. The dot row extends far past the actual key length, making it hard to tell where the field ends and awkward to click into and clear the field when rotating keys.

### Impact

- Visually confusing — the field looks much longer than the real key
- Makes it hard to select/clear the field when you want to change the key

---

mvanhorn:

---

## Why

After an LLM API key is saved, re-opening the LLM settings panel rendered a masked password field whose dot count matched the length of the stored key, stretching the row well past a reasonable width and making the field hard to select or clear. The legacy `settings-form.tsx` already seeds an empty value with a `"<hidden>"` placeholder when a key is set, but the newer LLM-profiles edit flow in `llm-settings-local-view.tsx` seeded the input with the actual stored value.

## Summary

- In `llm-settings-local-view.tsx`, seed the `llm.api_key` field empty when a key is already set (instead of `initialValues["llm.api_key"] = config.api_key`) and surface a fixed-length placeholder indicating a key is present.
- Preserve the existing submit logic so a blank field still means "keep the saved key" rather than "clear it."

## Issue Number

Fixes #1078

## How to Test

Run `npm install` then start the dev build. Save an LLM profile with an API key, reopen the settings panel, and confirm the key field shows a fixed-width placeholder rather than a dot row matching the key length. Submit with the field left blank and confirm the stored key is kept; enter a new value and confirm it replaces the stored key on submit.

## Video/Screenshots

N/A. Behavior change is covered by the unit tests listed in Notes; no UI recording was captured.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Covered by updated cases in `llm-settings-local-view.test.tsx`: edit mode shows the placeholder-masked field, blank submit keeps the key, a new value replaces it, and create mode shows an empty field with no misleading dots.

### DIFF
--- a/__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx
+++ b/__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx
@@ -388,8 +388,8 @@ describe("LlmSettingsLocalView", () => {
       });
 
       // The LlmSettingsScreen component receives initialValueOverrides
-      // with the profile's LLM config values. We verify this by checking
-      // that getProfile was called and the form is in edit mode.
+      // with the profile's non-secret LLM config values. The saved API key is
+      // intentionally represented by an empty field plus a hidden placeholder.
       expect(ProfilesService.getProfile).toHaveBeenCalledWith(
         "gpt-4-profile",
         "encrypted",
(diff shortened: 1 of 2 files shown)


neubig:

Issue #1098, opened 2026-06-03
[Bug]: Missing Runtime Services information in non development modes (NPM&Docker) seems to cause Automation creation to fail


DevinVinson:
### Operating System

macOS

### Installation Method

Docker (`ghcr.io/openhands/agent-canvas`)

### Agent Canvas Version

1.0.0-beta.6

### Bug Description

My agent was struggling to make an automation and was asking for an OPENHANDS_API_KEY. Digging deeper I saw that it had no Skills or `<RUNTIME_SERVICES>`.

If I compare my running dev copy to my docker I see that docker seems to cut off right at `</PROCESS_MANAGEMENT>` . The IMPORTANT, and CURRENT_DATETIME are also not included when compared to my `npm run dev` agent canvas.

### Steps to Reproduce

1. Run with docker and put a non-default local folder in the $PROJECTS_PATH value
2. Start a conversation: flip a coin
3. Click the + symbol, click Show Agent Tools & Metadata
4. Scroll to the bottom of the System Message and see content end with `</PROCESS_MANAGEMENT>`

### Actual Behavior

No skills loaded, no local runtime info.

### Expected Behavior

Everything should load.

### Relevant Logs

```shell

```

### Additional Context

I had OpenHands investigate and there is a proposed fix but with the missing skills and other content I am not going to try and make a PR just yet. Here is the results from that initial look:
```
 dug into where <RUNTIME_SERVICES> is set and why it can be missing in Docker. Here’s what I found and fixed:

✅ Root cause
<RUNTIME_SERVICES> is built from VITE_RUNTIME_SERVICES_INFO at build time (via Vite). The Docker image builds the frontend without that env var, so the runtime services block never gets appended to the agent’s system prompt.

✅ Fix implemented
I added a runtime injection path so the Docker entrypoint can supply the JSON at container startup:

scripts/static-server.mjs now accepts --runtime-services-info <json> and injects it into index.html as:
window.__AGENT_CANVAS_RUNTIME_SERVICES_INFO__
src/api/agent-server-adapter.ts now falls back to that window value when VITE_RUNTIME_SERVICES_INFO is missing.
docker/entrypoint.sh now builds the runtime services JSON (based on actual ports) and passes it into the static server.
Added a test case proving the window fallback works.
Files changed
src/api/agent-server-adapter.ts
scripts/static-server.mjs
docker/entrypoint.sh
__tests__/api/agent-server-adapter.test.ts
Notes
[...]

PR #1125, opened 2026-06-04
fix: populate <RUNTIME_SERVICES> in static builds so automations work

VascoSch92:
## Problem

Reported in #1098: in Docker (and the npm/binary install) the agent fails to create automations — it asks for an `OPENHANDS_API_KEY` or falls back to the cloud service (`app.all-hands.dev`), which rejects the local key.

## Root cause

The agent's `<RUNTIME_SERVICES>` system-prompt block (which tells the agent how to reach the local automation backend) is built from `VITE_RUNTIME_SERVICES_INFO`. That env var is only set **at build time** by the dev launchers (`scripts/dev-*.mjs` / `static-build.mjs`). Static builds run plain `npm run build` (`docker/Dockerfile`), so the var is empty, `parseRuntimeServicesInfo()` returns `null`, and the block is omitted.

Crucially, the dev trick **can't** be reused at `docker build` time: the ports/URLs (`PORT`, `AGENT_SERVER_URL`, `AUTOMATION_BASE_URL`, …) are runtime config, overridable at `docker run`, and the URLs are written from the *agent's* point of view (container loopback), not the browser's. So the info has to be supplied at container start, not baked into the image.

## Fix

Inject the info **at serve time**, mirroring the existing `--session-api-key` / `--auth-required` injection already used for the published binary:

- **`scripts/runtime-services-info.mjs`** (new) — single source of truth for the block's shape. Holds `buildRuntimeServicesInfo()` (moved out of `dev-safe.mjs`, which re-exports it for back-compat), extended with optional full-URL overrides, plus a small CLI so the entrypoint can emit the JSON with the *same* builder the dev stack uses.
- **`scripts/static-server.mjs`** — new `--runtime-services-info <json>` flag injects the JSON as `window.__AGENT_CANVAS_RUNTIME_SERVICES_INFO__` (dumb injector; doesn't know the shape).
- **`src/api/agent-server-adapter.ts`** — `parseRuntimeServicesInfo()` falls back to that window global when the env var is empty (same idiom as `getBakedSessionApiKey()`).
- **`docker/entrypoint.sh`** — runs the CLI with the sandbox-facing URLs it already exports (`AGENT_SERVER_URL`, `AUTOMATION_BASE_URL`, `127.0.0.1` to avoid IPv6 loopback) and passes the result to the static server(s). **`docker/Dockerfile`** ships the new module.

Responsibilities stay clean: one builder knows the shape, static-server is a dumb injector, the adapter is a dumb consumer.

## Scope

This fixes the missing `<RUNTIME_SERVICES>` block — the automation-creation failure that is the subject of #1098. The other observations in that issue (missing `IMPORTANT` / `CURRENT_DATETIME` sections and skills not loading) originate in the agent-server/SDK base prompt, not the frontend, and are **out of scope** here.

## Testing

- `npm run lint` (typecheck + eslint + prettier on `src`) — clean
- Full `vitest run` — 2977 passing (incl. new `runtime-services-info.test.ts`)
- New unit tests: window-fallback + env precedence (`agent-server-adapter.test.ts`); `--runtime-services-info` parse + injection (`static-server.test.ts`); URL-override builder + CLI end-to-end (`runtime-services-info.test.ts`)
- Validated the entrypoint CLI emits a `RuntimeServicesInfo`-shaped object that round-trips through the browser injection → `JSON.parse`

## Not yet validated

A live Docker build + automation-creation smoke test (logic, parsing, and round-trip are validated; an actual container run is not).

---

| Component | Value |
|---|---|
| **Automation** | `openhands-automation==1.0.0a6` |
| **Commit** | `20cbef97e46f0705c8c4fc53ae84aa68ac502b9f` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-20cbef9
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-20cbef9
```

```
ghcr.io/openhands/agent-canvas:sha-20cbef9-amd64
ghcr.io/openhands/agent-canvas:fix-runtime-services-info-static-builds-amd64
ghcr.io/openhands/agent-canvas:pr-1125-amd64
ghcr.io/openhands/agent-canvas:sha-20cbef9-arm64
ghcr.io/openhands/agent-canvas:fix-runtime-services-info-static-builds-arm64
ghcr.io/openhands/agent-canvas:pr-1125-arm64
ghcr.io/openhands/agent-canvas:sha-20cbef9
ghcr.io/openhands/agent-canvas:fix-runtime-services-info-static-builds
ghcr.io/openhands/agent-canvas:pr-1125
```

- Each tag (e.g., `sha-20cbef9`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-20cbef9-amd64`) are also available if needed


malhotra5 (approved):
LGTM

malhotra5:
@openhands can you modify the mock llm e2e tests and confirm that the runtime info is correctly constructed and included for conversations. maybe for the automations e2e test add this additional check

malhotra5:
The e2e test failure is unrelated to this PR so I'm going to approve it

PR #1136, opened 2026-06-04
fix: avoid invalid onboarding model defaults for OpenHands

### LINKED ISSUE #1111: [Bug]: Startup flow keeps invalid default model (claude-opus-4-5) and fails first test prompt
### Operating System
macOS

### Installation Method
npm (`npx @openhands/agent-canvas` or `npm install -g @openhands/agent-canvas`)

### Agent Canvas Version
1.0.0-alpha.10 - 1.0.0-beta.9

### Bug Description
During first-run startup/onboarding, selecting `OpenHands` as the provider and keeping the default model causes the test prompt to fail with a LiteLLM invalid-model error, even when the API key is correct.

This creates a broken first-run experience because users can proceed with a default model that is not actually available for their key.

### Steps to Reproduce
1. Launch Agent Canvas fresh and go through startup onboarding.
2. In LLM setup, choose provider `OpenHands`.
3. Leave the default model unchanged (`claude-opus-4-5`).
4. Enter a valid API key.
5. Click the onboarding test/submit prompt (example: ask it to create an HTML file).

### Actual Behavior
The request fails with:

```text
litellm.BadRequestError: Litellm_proxyException - /chat/completions: Invalid model name passed in model=claude-opus-4-5. Call `/v1/models` to view available models for your key.
```

### Expected Behavior
Onboarding should not allow the first test prompt to run with a model that is invalid for the provided key.

Expected behavior options (any of these would resolve the issue):
- Validate model availability against the entered key before allowing submit.
- Auto-select a model known to be available for that key.
- If unavailable, present a guided model selection instead of surfacing the raw backend error.

### Relevant Logs
```text
litellm.BadRequestError: Litellm_proxyException - /chat/completions: Invalid model name passed in model=claude-opus-4-5. Call `/v1/models` to view available models for your key.
```

### Additional Context
- This appears key/model-entitlement related, not an API key typo.
- In this case, the default alias model (`claude-opus-4-5`) was not accepted for the key, while other model IDs were available.
- This regression is especially painful in first-run onboarding because users interpret it as setup failure even with a valid key.

FraterCCCLXIII:
## Summary
- switch the OpenHands onboarding default model to an Anthropic model to avoid invalid first-run defaults
- add/keep onboarding coverage that asserts the OpenHands path seeds the Anthropic default model value
- show `Select Model` as the placeholder in the empty model dropdown to make model selection state explicit

Fixes #1111

## Test plan
- [x] `npm test -- __tests__/components/onboarding/onboarding-modal.test.tsx`
- [x] `npm test -- __tests__/components/modals/settings/model-selector.test.tsx`

Made with [Cursor](https://cursor.com)


neubig:

PR #1134, opened 2026-06-04
Preserve LLM profile base URL in Basic settings

xingyaoww:
## Summary

Keep the existing Basic-tab OpenHands behavior: when the model is `openhands/*`, save the All-Hands LiteLLM proxy `base_url` explicitly.

For every other provider, stop deleting `base_url` on Basic-tab save. The Basic form does not show `base_url`, so an existing custom endpoint should be preserved rather than silently removed.

## Validation

- `npm test -- __tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx`
- `npm run typecheck`

---
**🐳 Docker images for this PR**

• **GHCR package:** https://github.com/OpenHands/agent-canvas/pkgs/container/agent-canvas

| Component | Value |
|---|---|
| **Image** | `ghcr.io/openhands/agent-canvas` |
| **Architectures** | amd64, arm64 |
| **Agent Server** | `ghcr.io/openhands/agent-server:1.24.0-python` |
| **Automation** | `openhands-automation==1.0.0a6` |
| **Commit** | `4e0363e8c960e2d37d753a9fafd5fc62a28ac483` |

**Pull (multi-arch manifest)**
```bash
# Multi-arch manifest — Docker automatically pulls the correct architecture
docker pull ghcr.io/openhands/agent-canvas:sha-4e0363e
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-4e0363e
```

**All tags pushed for this build**
```
ghcr.io/openhands/agent-canvas:sha-4e0363e-amd64
ghcr.io/openhands/agent-canvas:codex-fix-llm-profile-proxy-base-url-amd64
ghcr.io/openhands/agent-canvas:pr-1134-amd64
ghcr.io/openhands/agent-canvas:sha-4e0363e-arm64
ghcr.io/openhands/agent-canvas:codex-fix-llm-profile-proxy-base-url-arm64
ghcr.io/openhands/agent-canvas:pr-1134-arm64
ghcr.io/openhands/agent-canvas:sha-4e0363e
ghcr.io/openhands/agent-canvas:codex-fix-llm-profile-proxy-base-url
ghcr.io/openhands/agent-canvas:pr-1134
```

**About Multi-Architecture Support**
- Each tag (e.g., `sha-4e0363e`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Docker automatically pulls the correct architecture for your platform
[...]

PR #1148, opened 2026-06-04
fix: preserve OpenHands proxy base_url for rewritten litellm_proxy profiles

### LINKED ISSUE #1146: [Bug]: LLM Profiles are not working well
### Operating System

macOS

### Installation Method

npm (`npx @openhands/agent-canvas` or `npm install -g @openhands/agent-canvas`)

### Agent Canvas Version

several versions beta.6 and after

### Bug Description

Sorry for the vague bug!

We're getting various reports of LLM profiles not working well:
* First time setup, with OpenHands agent, and a claude code token often (not always!) triggers a litellm auth error
* First time setup is often plugging in OpenHands as the provider
* Deleting the active LLM profile doesn't set the next one as active, leaving you in a weird state where convos don't have an LLM
* LLM profiles randomly switch when starting a conversation

Many of these are not consistently reproducible.

### Steps to Reproduce

I'd suggest repeatedly clearing out all state (localstorage and .openhands) then doing onboarding, to try and trigger the problems.

### Actual Behavior

chaotic

### Expected Behavior

predictable

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

VascoSch92:
## Summary

Fixes #1146 — agent-canvas could persist an LLM profile in the invalid shape `model: "litellm_proxy/*"` + `base_url: null`, after which the first LLM call fails with a misleading OpenAI auth error (`Incorrect API key provided: sk-...`) even though an OpenHands provider key is in use.

## Root cause

OpenHands provider models are submitted by the GUI as `openhands/<model>`. On save, the agent-server's LLM validator **rewrites** them to `litellm_proxy/<model>` paired with the All-Hands proxy `base_url`. So once a profile has been saved once, its stored model is `litellm_proxy/*`.

The "is this an OpenHands model?" check (`isOpenHandsProviderModel`) only matched the `openhands/*` prefix. Every Basic-view save path re-derives `base_url` from the model, and for a stored (`litellm_proxy/*`) OpenHands profile that check returned **false** — so the proxy `base_url` was dropped/nulled:

- `src/components/features/settings/llm-profiles/llm-settings-local-view.tsx` (local profile editor) — `delete llmConfig.base_url`
- `src/routes/llm-settings.tsx` `buildPayload` (standalone / onboarding / cloud screen) — reset to the schema default (`null`)

With the proxy `base_url` gone, `litellm_proxy/*` has no `api_base`, so LiteLLM routes to the default OpenAI endpoint → the OpenHands key is rejected. The UI also can't normalize the model back to the "OpenHands" provider display, because that normalization keys off the known proxy `base_url`.

This requires a round-trip (save → reload → Basic save again), which is why a clean onboarding looks fine but a reused profile breaks — matching the report.

## Fix

- Add `isOpenHandsProxyModel(model, baseUrl)` (+ `isOpenHandsProxyBaseUrl`) in `src/utils/openhands-llm.ts`: true for `openhands/*` **or** the SDK-rewritten `litellm_proxy/*` when it is paired with the All-Hands proxy URL.
- Use it in both Basic-save paths so the proxy `base_url` is preserved (re-stamped) instead of dropped.
- `normalizeDisplayModel` now reuses the shared proxy-URL check rather than keeping its own duplicate.

## Tests

- `__tests__/utils/openhands-llm.test.ts` — full matrix for the new detectors, including negative cases (third-party `litellm_proxy` gateway, already-stranded `base_url:null`, plain providers paired with the proxy URL).
- `__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx` — regression using the real post-rewrite shape (`litellm_proxy/claude-opus-4-8` + proxy URL); a Basic-tab re-save keeps `base_url`. The suite previously only covered the transient `openhands/*` shape, which is how this slipped through.

`npm test`, `npm run typecheck`, and `eslint`/`prettier` on the changed `src` files all pass.

## Out of scope

For **non**-OpenHands providers the Basic tab still drops a custom `base_url` (it has no field to display it) — the separate concern from the earlier #1134. Left out here to keep this fix focused on the #1146 invariant.

---

| Component | Value |
|---|---|
| **Automation** | `openhands-automation==1.0.0a6` |
| **Commit** | `dd00f9c03e29647a1c485cdfb22e8f2881d018e1` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-dd00f9c
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-dd00f9c
```

```
ghcr.io/openhands/agent-canvas:sha-dd00f9c-amd64
ghcr.io/openhands/agent-canvas:fix-llm-profile-proxy-base-url-1146-amd64
ghcr.io/openhands/agent-canvas:pr-1148-amd64
ghcr.io/openhands/agent-canvas:sha-dd00f9c-arm64
ghcr.io/openhands/agent-canvas:fix-llm-profile-proxy-base-url-1146-arm64
ghcr.io/openhands/agent-canvas:pr-1148-arm64
ghcr.io/openhands/agent-canvas:sha-dd00f9c
ghcr.io/openhands/agent-canvas:fix-llm-profile-proxy-base-url-1146
ghcr.io/openhands/agent-canvas:pr-1148
```

- Each tag (e.g., `sha-dd00f9c`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-dd00f9c-amd64`) are also available if needed


enyst:
I wonder, could we fix it in the SDK instead, or am I missing something? Why is this happening:
> OpenHands provider models are submitted by the GUI as openhands/<model>. On save, the agent-server's LLM validator rewrites them to litellm_proxy/<model> paired with the All-Hands proxy base_url. So once a profile has been saved once, its stored model is litellm_proxy/*.

@openhands Find the code referenced here and tell me the PR that added it, then investigate if we can KEEP `openhands/` provider stored, over the wire, etc, and only translate it when we send the request to litellm?

enyst:
I checked `OpenHands/software-agent-sdk`.

**Where the rewrite lives today**

Current SDK code is here: [`openhands-sdk/openhands/sdk/llm/llm.py#L499-L520`](https://github.com/OpenHands/software-agent-sdk/blob/main/openhands-sdk/openhands/sdk/llm/llm.py#L499-L520). It is a Pydantic `@model_validator(mode="before")` on `LLM`:

```py
if model_val.startswith("openhands/"):
    model_name = model_val.removeprefix("openhands/")
    d["model"] = f"litellm_proxy/{model_name}"
    d["base_url"] = d.get("base_url") or "https://llm-proxy.app.all-hands.dev/"
```

**PR archaeology**

The behavior was originally added in `openhands/core/config/llm_config.py` by OpenHands/software-agent-sdk#70 (`f16a2da9`, "Add remaining LLM fixes and LLM registry"). It was later moved into the current `LLM` pydantic class by OpenHands/software-agent-sdk#91. Related base-url tweaks happened in OpenHands/software-agent-sdk#338 and OpenHands/software-agent-sdk#1341.

**Why it is persisted / returned over the wire**

This is not only a LiteLLM request-time mapping. Because the rewrite runs during `LLM` validation, the public `LLM.model` field is already mutated before agent-server handlers see it:

- `profiles_router.SaveProfileRequest` has `llm: LLM`, so FastAPI validation rewrites the request body before `save_profile()` runs.
- `LLMProfileStore.save()` writes `llm.to_persisted()`, which `model_dump()`s the already-rewritten model.
- `GET /api/profiles/{name}` loads the profile as `LLM` and returns `llm.model_dump()`.
- settings have the same shape because `OpenHandsAgentSettings.llm` is also an `LLM`, and `PATCH /api/settings` returns `settings.agent_settings.model_dump()`.

So after one round trip, the UI can legitimately receive `model: "litellm_proxy/..."` for a profile the user selected as OpenHands.

**Can we keep `openhands/` stored/on the wire and translate only when calling LiteLLM?**

Yes, I think that is feasible and probably the cleaner SDK-level design, but it is a behavior/API change in the SDK rather than a one-line validator fix. The SDK would need to separate the public model id from the LiteLLM transport model, roughly:

1. Stop mutating `d["model"]` in the validator; only default `base_url` for `openhands/*`.
2. Add a transport-model helper (`openhands/foo` -> `litellm_proxy/foo`) and use it in `_prepare_transport_kwargs()` before calling `litellm.completion` / `acompletion` / responses.
3. Use that transport model anywhere the SDK currently depends on the rewritten value, especially provider inference, LiteLLM proxy model-info lookup, and capability/feature lookups where needed.
4. Add profile/settings migrations or normalization for old persisted profiles: `litellm_proxy/<known OpenHands model>` + the OpenHands proxy base URL should round-trip back to `openhands/<model>`. `LLMProfileStore.list_summaries()` needs special attention because it currently reads raw JSON without constructing an `LLM`.
5. Keep third-party LiteLLM proxy configs safe: do not convert arbitrary `litellm_proxy/*` unless it is paired with the known OpenHands proxy URL (and probably a verified OpenHands model). If the profile is already stranded as `litellm_proxy/*` + `base_url: null`, the SDK cannot prove it was OpenHands without a heuristic.

**Recommendation for this PR**

I agree the long-term home for this normalization is the SDK. I would still keep this agent-canvas fix for now because released/current agent-server versions already return and persist the rewritten shape, and users may already have profiles in that shape. This PR is a compatibility guard: when the current API returns `litellm_proxy/*` + the known OpenHands proxy URL, the Basic save path must not drop that URL and create the invalid `litellm_proxy/* + base_url:null` state.

A follow-up SDK PR could then make `openhands/*` the stable stored/API model and translate only at transport time; once agent-canvas pins an agent-server version with that migration, we can simplify the frontend workaround.

_This comment was generated by an AI agent (OpenHands) on behalf of the PR author._

__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx
```diff
@@ -582,6 +582,49 @@ describe("LlmSettingsLocalView", () => {
...
       expect(savedLlm.base_url).toBe(OPENHANDS_LLM_PROXY_BASE_URL);
     });
+
+    it("keeps the proxy base_url for a stored litellm_proxy OpenHands model (issue #1146)", async () => {
+      // Arrange — the shape actually persisted after onboarding through the
+      // OpenHands provider: the SDK has already rewritten openhands/* to
+      // litellm_proxy/* and paired it with the All-Hands proxy base URL. A
+      // Basic-tab re-save must not strip that base URL, otherwise the profile
+      // is stranded as `litellm_proxy/* + base_url:null` and LiteLLM reroutes
+      // the OpenHands key to the default OpenAI endpoint.
```
enyst, line 592: @openhands but when the user has a DIFFERENT base url, you know, anything, say openrouter base url, then they go to basic settings, do we behave consistently with `openhands/` provider behavior AFTER this PR ? 

Respond directly on github in reply here.

enyst: Yes for the `openhands/*` provider path: after this PR, Basic save treats `openhands/*` as OpenHands-backed regardless of whatever stale/different `base_url` was present, and re-stamps it to `https://llm-proxy.app.all-hands.dev/`. That is covered by the test immediately above this one (`openhands/claude...` + `https://stale.example.com/v1` -> proxy URL).

For `litellm_proxy/*` with a *different* base URL, e.g. an OpenRouter/custom LiteLLM gateway, we intentionally do **not** classify it as OpenHands-backed. In Basic view it follows the existing non-OpenHands behavior: there is no base-url field, so the save path drops/resets `base_url` rather than preserving the custom gateway. That is the pre-existing #1134 concern and I left it out of this PR on purpose to avoid hijacking third-party `litellm_proxy/*` profiles into the All-Hands proxy.

So: OpenHands provider behavior is consistent for `openhands/*` and for the SDK-rewritten `litellm_proxy/*` only when it is paired with the known All-Hands proxy URL. Custom/OpenRouter base URLs are unchanged/out of scope here.

_This reply was generated by an AI agent (OpenHands) on behalf of the PR author._

enyst: This is to say it’s _inconsistent_.

enyst (approved):

enyst:
@VascoSch92 TBH, I think this may be necessary right now, but it’s a bit scary to me that we keep adding code, behaving inconsistently with other providers even, adding special cases for a simple issue: we just need an `openhands/` provider which is unknown to LiteLLM.
I do think the best solution is to only rewrite it just before we send calls to LiteLLM, then all these special cases go away. If I’m wrong, please let me know!

On a different note, it’s not totally clear to me why would people save the custom LLM in Basic View… 🤔 but I approved the PR! MVP and all. 😅

PR #1168, opened 2026-06-05
Add frontend performance fixes for long conversations and load times

DevinVinson:

---

## Summary
Improve long-conversation performance and retained memory behavior in the conversation view.

### Rendering lookup optimization

Conversation event rendering repeatedly needs to match observations back to their producing actions. The chat list now builds one `Map<eventId, ActionEvent>` from the full event history and passes it through the render/grouping path, avoiding repeated full-history `Array.find` scans while preserving compatibility fallbacks for direct callers and tests.

### Retained event payload sanitization

Browser observations can carry large base64 screenshots. Those screenshots are now stripped before events are retained in React Query/Zustand history, while the live WebSocket event remains available long enough to update the Browser tab. This reduces retained memory growth from browser-heavy conversations without changing visible chat history.

### Terminal replay retention

The terminal replay store now keeps only the latest retained entries and truncates oversized output chunks with an explicit marker. This bounds the common terminal-output memory path while preserving recent replay context.

### Remote-history render warning fix

Moved user-message truncation parent notification out of layout measurement and into an effect. This avoids React’s cross-component update warning when older messages load slowly from a remote/cloud backend.

### Documentation

Added `docs/conversation-performance.md` with invariants for action lookup, retained payload sanitization, and terminal replay retention.

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/docs/conversation-performance.md
+++ b/docs/conversation-performance.md
@@ -1,0 +1,29 @@
+# Conversation performance notes
+
+The conversation view is the highest-risk UI surface for unbounded growth:
+long-running sessions accumulate events, tool outputs, browser screenshots, and
+rendered DOM. This document records the current performance guardrails so future
+changes can preserve the same invariants.
+
+## Event rendering action lookup
+
+Conversation rendering frequently needs to pair an observation with the action
+that produced it. The message list, event grouping, and group summary all use
+that relationship to render titles, thoughts, reasoning content, and tool
+details.
+
+The render path now builds a single `Map<eventId, ActionEvent>` for the current
+event history in `Messages`, then passes that lookup to children. This avoids
+repeated `Array.find` scans over the full conversation history for every
+observation or group. The compatibility fallback still accepts a raw event list
+for direct component tests and non-list callers, but production rendering should
+prefer the precomputed lookup.
+
+Invariants:
+
+- Build the lookup from the full event history, not only the UI-reduced event
+  list, because observations may need actions that have been replaced in
+  `uiEvents`.
+- Keep the lookup scoped to a render of `Messages`; it should not become a
+  long-lived global cache.
+- Do not reintroduce per-observation full-history scans in hot render paths.
--- a/docs/conversation-performance.md
+++ b/docs/conversation-performance.md
@@ -49,3 +49,24 @@ Invariants:
   links, or visible text.
 - If another event type starts carrying large binary or base64 data, add it to
   the same sanitizer rather than handling it in one fetch path only.
+
+## Terminal replay retention
+
+The Terminal tab replays command input and output from `useCommandStore` when
+the xterm instance remounts. Without retention bounds, long-running
+conversations can keep a large command array and very large output strings alive
+even when the terminal is not visible.
+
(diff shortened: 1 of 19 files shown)


docs/conversation-performance.md
```diff
@@ -0,0 +1,72 @@
+# Conversation performance notes
+
+The conversation view is the highest-risk UI surface for unbounded growth:
+long-running sessions accumulate events, tool outputs, browser screenshots, and
+rendered DOM. This document records the current performance guardrails so future
```
enyst, line 5: Your idea of a timeline is even more tempting. Sure, it would have this problem if not addressed, but I have hopes for clearing the regular view for people, because it can form habits, and we might discover that the timeline view will not be used every conversation 🤔

docs/conversation-performance.md
```diff
@@ -0,0 +1,72 @@
...
+The render path now builds a single `Map<eventId, ActionEvent>` for the current
+event history in `Messages`, then passes that lookup to children. This avoids
+repeated `Array.find` scans over the full conversation history for every
+observation or group. The compatibility fallback still accepts a raw event list
+for direct component tests and non-list callers, but production rendering should
+prefer the precomputed lookup.
+
+Invariants:
+
+- Build the lookup from the full event history, not only the UI-reduced event
```
enyst, line 24: This is not going to work IMHO, if I understand correctly. In a bit crazy experiment for its time, at the end of 2025 (a lifetime ago! 😓 ), I had a conversation of **60k events**. I think maybe we should not even try to hold those in memory to build anything...

src/components/conversation-events/chat/group-events.ts
```diff
@@ -59,18 +63,23 @@ export type RenderedItem =
  * the event itself. This keeps reasoning text in the main message stream
  * instead of buried inside a collapsed action group.
  *
- * `allEvents` should be the full event history so observations can find
- * their matching action; if omitted, it defaults to `events`.
+ * `actionLookupOrEvents` should be an action lookup built from the full event
+ * history so observations can find their matching action in O(1). Passing the
```
enyst, line 67: Aha, so this is where this idea was coming from - the deleted text seeded the idea

It's not accurate in the deleted text, and so I think maybe we don't need to assume it

enyst (commented):
Just want to express a thought, for clarity: I think the maximum number of events that are necessary is `max_size` or close to. `max_size` is the number of events at which a condensation happens for sure.

When condensed, the events that go to the LLM are only:
- system prompt
- (currently also) initial user message
- summary of older events (from condensation)
- the most recent `max_size / 2` events.

Not full history, not a single event above `max_size`, not older events than the condensation summary event.

They don't exist for the LLM. So do they NEED to exist for the user? Maybe, but the older events are truly irrelevant now, and IMHO maybe we will be happier to always assume and implement and document everywhere that "events" means normally maximum above.

Then, in a timeline view, for example, we can do pagination and give the user optional access to older events... 🤔 WDYT?

Thank you for this!

PR #1178, opened 2026-06-05
fix: single-line agent-server logs in dev:automation mode

tofarr:

---

## Why

In `npm run dev` / `dev:automation` mode, agent-server log lines were hard to read for two reasons:

1. **Duplicate timestamp** — Python's `RichHandler` prepends its own `[HH:MM:SS]` column on top of the `HH:MM:SS [agent-server]` prefix that `logService` already adds.
2. **Line wrapping** — Rich wraps long log messages (e.g. HTTP access log URLs) across multiple lines based on terminal width. `spawnService` splits piped output on `\n`, so each continuation line gets its own `logService` prefix, turning one log record into 2–3 entries.

## Summary

- Set `LOG_JSON=true` for the agent-server process in `dev-with-automation.mjs` so the SDK emits one compact JSON object per log record (no Rich formatting, no wrapping, no duplicate timestamp).
- Add `parseAgentServerLogLine()` which parses each JSON line and re-formats it as a single human-readable string: `LEVEL     <message>  filename.py:lineno`.
- Add a `parseLogLine` option to `spawnService`

## Screenshots

**Before**
image

**After**
image

### DIFF
--- a/scripts/dev-with-automation.mjs
+++ b/scripts/dev-with-automation.mjs
@@ -119,6 +119,38 @@ function logError(message) {
   console.error(`${c.red}✗${c.reset} ${message}`);
 }
 
+/**
+ * Parse one JSON log line produced by the SDK's JsonFormatter and return a
+ * single-line human-readable string + an appropriate ANSI color.
+ *
+ * Returns null for non-JSON lines so callers can fall back to the raw text.
+ *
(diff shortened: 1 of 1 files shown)


hieptl (approved):
Thank you! 🙏

Issue #233, opened 2026-05-09
Design a maintainable remote backend bootstrap flow for Agent Canvas


xingyaoww:
## Background

The current `Add Backend with Agent` concept starts a local agent conversation that helps the user configure another machine as an Agent Canvas backend. The first prototype proves the UI entry point and conversation-start flow, but the setup prompt is already carrying too much procedural detail.

We should make this easier to understand and maintain by moving the remote-machine setup contract into versioned scripts and metadata files in this repo, instead of relying on a long prompt.

## Problem

A prompt-only flow has a few issues:

- Users cannot quickly understand what will happen on the target machine.
- The agent may scatter files or invent slightly different setup commands across runs.
- Remote setup state is hard to inspect, update, and repair later.
- The UX needs to support both cases: the agent can SSH to the target machine, or the user must run a command manually on that target.

## Proposed direction

Use a standard remote bootstrap contract:

1. Keep the UI entry point in the backend selector as `Add Backend with Agent`.
2. Start a new conversation using the currently configured LLM profile/model.
3. Keep the initial prompt short. It should ask:
   - What target machine should be configured?
   - Is SSH from this local machine already configured?
   - Are the local and target machines on the same LAN/private network?
   - Does the user authorize the agent to connect over SSH and perform setup?
   - Should this be temporary/dev or long-running?
4. Move the detailed implementation into a repo-owned bootstrap script, for example a future `scripts/remote-bootstrap.sh` or package/CLI equivalent.
5. Standardize target-machine state under `~/.openhands/agent-canvas`:
   - `repo/` for the agent-canvas checkout
   - `env/.env` for stable runtime secrets, with restrictive permissions
   - `bin/agent-canvas-stack` for idempotent maintenance commands
   - `metadata/backend.json` for non-secret machine/backend metadata
   - `metadata/connection.md` for local UI connection instructions
   - `metadata/setup-log.md` for setup/update decisions
6. The fixed maintenance command should support at least:
   - `setup`
   - `start`
   - `stop`
   - `status`
   - `logs`
   - `update`
   - `print-connection`

## Manual-install fallback

If the agent cannot SSH to the target machine, it should give the user a single command to run on the target machine. Ideally this is eventually something like:

```bash
curl -fsSL https://raw.githubusercontent.com/OpenHands/agent-canvas/main/scripts/remote-bootstrap.sh | bash
```

or, if we publish a CLI/package later:

```bash
npx @openhands/agent-canvas remote-bootstrap
[...]

Issue #68, opened 2026-05-01
Use backend-persisted settings and secrets instead of localStorage


neubig:
## Summary

Move `agent-server-gui`'s real settings/secrets persistence from browser `localStorage` to the selected backend's settings/secrets APIs.

Only backend connection bootstrap data should remain in `localStorage`; runnable agent configuration and credentials should be backend-owned.

## Current behavior

* `src/api/settings-service/settings-service.api.ts` reads/writes `localStorage` key `openhands-agent-server-settings`.
* `src/api/secrets-service.ts` stores git provider tokens in `localStorage` key `openhands-agent-server-git-provider-tokens`.
* `src/api/agent-server-config.ts` stores backend URL/session key in `localStorage` key `openhands-agent-server-config`.

The last item should remain local. The first two should move to backend APIs.

## Dependencies

Depends on backend support in OpenHands/software-agent-sdk:

* OpenHands/software-agent-sdk#3034 / PR [#3026](<https://github.com/OpenHands/agent-server-gui/issues/3026>) for backend-persisted settings/secrets.
* OpenHands/software-agent-sdk#3035 for LLM profile and git-provider credential APIs.

### Keep in browser localStorage

* backend URL/session key/working-dir bootstrap config;
* selected backend id in future multi-backend work;
* UI-only state such as drafts, tab state, dismissed banners.

### Move to backend

* LLM settings/API key metadata;
* agent settings and conversation settings;
* MCP config;
* disabled skills;
* search API key metadata/secret value;
* git provider tokens and `provider_tokens_set` metadata;
* custom secrets;
* LLM profiles once supported by backend.

### SettingsService

* `getSettings()` calls the selected backend settings endpoint instead of reading `openhands-agent-server-settings`.
* `saveSettings()` calls the backend update endpoint using the canonical partial-update shape, expected to include `agent_settings_diff` and `conversation_settings_diff`.
* Schema reads still use backend schema endpoints.
* Secret fields follow backend redaction conventions (`*_set` booleans / `api_key_set`) and must not round-trip mask strings as real secrets.

### SecretsService

* Custom secrets use backend CRUD endpoints.
* Git provider tokens use backend git-provider credential endpoints.
* Remove writes to `openhands-agent-server-git-provider-tokens`.
* `provider_tokens_set` comes from backend settings responses.

### Legacy localStorage import

If legacy keys exist after connecting to a compatible backend:

[...]

PR #1208, opened 2026-06-07
feat(backend-registry): seed deployment-provided default backends

shanemort1982:
## Problem

The backend registry lives entirely in each browser's `localStorage` (`openhands-backends`). For a **self-hosted, multi-user / multi-device** Agent Canvas this is painful:

- If one person adds a backend, nobody else sees it — every teammate must re-add it by hand.
- The same user has to re-add backends on every browser/device (office, home, laptop, phone).
- There is currently no way for a deployment to ship a known fleet of backends to all of its users.

This is the gap behind the "Keep in browser localStorage … selected backend id" note in #68 and the bootstrap discussion in #233: today there's no server-owned/deployment-owned source for the backend list.

## What this does

Lets a deployment advertise a fixed list of default backends that are auto-registered in **every** browser, while leaving per-user customisation intact.

- **New** `getDeploymentDefaultBackends()` (`src/api/backend-registry/deployment-backends.ts`) reads a backend list from, in order:
  1. `import.meta.env.VITE_DEFAULT_BACKENDS` (baked at build time, handy for `npm run dev`), then
  2. the injected `window.__AGENT_CANVAS_DEFAULT_BACKENDS__` global (serve time).

  This mirrors exactly how `getBakedSessionApiKey()` sources the session key. Entries only need `{host, apiKey}`; `name` defaults to the host, `kind` to `"local"`, and `id` to a stable `deployment:<host>` value (so merges are idempotent and the same backend gets the same id across browsers). Invalid entries are dropped and the list is de-duplicated by host.

- `readStoredBackends()` (`storage.ts`) now **merges** these defaults into the registry **by host**:
  - seeds them on first load (even when no launcher local backend is available), and
  - appends any that are missing on subsequent reads,
  - **never** removing or overwriting user-added/edited entries (a default whose host already exists is skipped, so users can still rename/edit their local copy).

- `scripts/static-server.mjs` gains `--default-backends <json>` (and the `AGENT_CANVAS_DEFAULT_BACKENDS` env var). The value is validated (must be a JSON array) and injected as a clean JS array literal into `index.html`, alongside the existing session-key / runtime-services injection.

## Usage

```bash
# published binary / Docker entrypoint
node scripts/static-server.mjs \
  --default-backends '[{"name":"prod (.238)","host":"http://192.168.1.238:8000","apiKey":"<session-key>"}]'

# or via env
AGENT_CANVAS_DEFAULT_BACKENDS='[{"host":"http://10.0.0.5:8000","apiKey":"…"}]' node scripts/static-server.mjs
```

## Tests

- `__tests__/api/backend-registry/deployment-backends.test.ts` — parsing (array + JSON string), env-over-window precedence, validation, de-dup, malformed JSON.
- `__tests__/api/backend-registry/storage.test.ts` — seed-on-empty, merge-into-existing (+ persistence), no-duplicate when host already registered.
- `__tests__/scripts/static-server.test.ts` — `parseArgs` for `--default-backends`, HTML injection of the global, non-array ignored, null = no injection.

```
npm run make-i18n && npx vitest run __tests__/api/backend-registry __tests__/scripts/static-server.test.ts   # 67 passed
npm run typecheck   # clean
eslint + prettier   # clean on changed files
```

## Notes / follow-ups

- Backwards compatible: with no env/flag/global set, behaviour is unchanged.
- Removal semantics: a deployment default that a user deletes will reappear on next load (it is, by design, deployment-owned). If per-user opt-out is desired, a `openhands-backends-removed` tombstone set could be a small follow-up.
- Relates to #68 and #233.

### DIFF
--- a/__tests__/api/backend-registry/deployment-backends.test.ts
+++ b/__tests__/api/backend-registry/deployment-backends.test.ts
@@ -1,0 +1,87 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { getDeploymentDefaultBackends } from "#/api/backend-registry/deployment-backends";
+
+type MutableWindow = Record<string, unknown>;
+
+function setInjectedBackends(value: unknown): void {
+  (window as unknown as MutableWindow).__AGENT_CANVAS_DEFAULT_BACKENDS__ =
+    value;
+}
+
+afterEach(() => {
+  delete (window as unknown as MutableWindow).__AGENT_CANVAS_DEFAULT_BACKENDS__;
+  vi.unstubAllEnvs();
+});
+
+describe("getDeploymentDefaultBackends", () => {
+  it("returns an empty list when nothing is configured", () => {
+    expect(getDeploymentDefaultBackends()).toEqual([]);
+  });
+
+  it("reads an injected window global (array form)", () => {
+    setInjectedBackends([
+      { host: "http://10.0.0.5:8000", apiKey: "key-a", name: "Box A" },
+    ]);
+
+    expect(getDeploymentDefaultBackends()).toEqual([
+      {
+        id: "deployment:http://10.0.0.5:8000",
+        name: "Box A",
+        host: "http://10.0.0.5:8000",
+        apiKey: "key-a",
+        kind: "local",
+      },
+    ]);
+  });
+
+  it("reads an injected window global (JSON string form)", () => {
+    setInjectedBackends(
+      JSON.stringify([{ host: "10.0.0.6:8000", apiKey: "key-b" }]),
+    );
+
+    const result = getDeploymentDefaultBackends();
+    expect(result).toHaveLength(1);
+    expect(result[0]).toMatchObject({
+      host: "http://10.0.0.6:8000",
+      apiKey: "key-b",
+      name: "http://10.0.0.6:8000",
+      kind: "local",
+    });
+  });
+
+  it("prefers the build-time env var over the window global", () => {
+    vi.stubEnv(
+      "VITE_DEFAULT_BACKENDS",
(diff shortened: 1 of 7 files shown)


shanemort1982:
Context for maintainers: this came out of dogfooding agent-canvas across a real self-hosted fleet (5 boxes on a LAN), after @enyst kindly suggested trying it over the old UI. 🙏

Two things surfaced in practice:

1. Running a **stock** `docker run ghcr.io/openhands/agent-server:<tag>-python` as a backend 500s on conversation start (`KeyError: ToolDefinition 'canvas_ui' is not registered`), because the frontend always injects `tool_module_qualnames.canvas_ui`. Write-up + an `OH_EXTRA_PYTHON_PATH` workaround in OpenHands/agent-canvas#2108.
2. The backend registry is per-browser `localStorage`, so multi-user / multi-device self-hosting means everyone re-adds the same backends. This PR lets a deployment seed default backends for every browser (relates to OpenHands/agent-canvas#68 and OpenHands/agent-canvas#2312).

Tests / typecheck / lint are green locally (67 passing). Since this is a fork PR, CI and the Vercel preview are waiting on a maintainer to approve — would appreciate someone kicking those off and taking a look when convenient. Happy to adjust scope (e.g. per-user opt-out tombstones) to match your preference.

_Disclosure: the investigation and this PR were done with the OpenHands agent on behalf of @shanemort1982._

shanemort1982:
The Vercel preview deployment is the only failing check on this PR — it errors with **"Authorization required to deploy."** That's a deploy-authorization gate, not a code problem: every other check (typecheck, eslint, prettier, i18n completeness, and the unit tests) is green.

Could a maintainer please authorise the Vercel preview deployment for this PR so the preview can build? Thanks.

neubig:

Issue #1220, opened 2026-06-07
[Bug] Inconsistent i18n key usage: raw strings vs I18nKey enum


VascoSch92:
Some components use the type-safe `I18nKey` enum:

```ts
t(I18nKey.COMMON$NEW_CONVERSATION)
```

While others use raw string literals:

```ts
t("COMMON$NEW_CONVERSATION")
t("DIFF_VIEWER$WAITING_FOR_RUNTIME")
```

This defeats the purpose of the generated enum — if a key is renamed or removed, string-literal usages won't produce compile-time errors.

**Suggestion:** Enforce `I18nKey` enum usage via an ESLint rule that disallows raw string arguments to `t()`.

PR #1230, opened 2026-06-08
fix: require I18nKey for translation calls

muff-c:

---

## Why

Raw string keys passed to `t()` do not fail typecheck when a translation key is renamed or removed.

## Summary

- Replaced raw translation-key strings passed to `t()` with `I18nKey`.
- Added an ESLint guard for new raw translation-key calls.
- Added the missing `SETTINGS$UPGRADE_BUTTON` translation key that was exposed by the change.

## Issue Number

Refs #1220

## How to Test

- `npm run lint`
- `npm run build`
- `npm run check-translation-completeness`
- `npm test -- --run __tests__/i18n/duplicate-keys.test.ts __tests__/i18n/files-diff-label.test.ts __tests__/i18n/sidebar-mcp-directory-label.test.ts`

## Video/Screenshots

No UI change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


VascoSch92:
@muff-c is that ready for review?

src/i18n/translation.json
```diff
@@ -22933,13 +22933,55 @@
     "uk": "Пісочницю цієї розмови видалено. Наведена вище історія доступна лише для читання."
   },
   "CHAT_INTERFACE$ACP_RESUME_SANDBOX_DESCRIPTION": {
-    "en": "This conversation's sandbox was recycled. Resume to continue where you left off."
+    "en": "This conversation's sandbox was recycled. Resume to continue where you left off.",
```
VascoSch92, line 23004: Why are you adding these translations? Where missing?

If yes, cool it means that the I18nKey is working :-)

VascoSch92 (commented):
I left a question... otherwise LGTM

src/i18n/translation.json
muff-c: Yep, exactly. `SETTINGS$UPGRADE_BUTTON` was already used as a raw key, and switching it to `I18nKey` made typecheck catch that it was missing from `translation.json`.

The ACP entries came from upstream with only `en`; after updating the branch, `check-translation-completeness` failed, so I filled the missing locale slots with the same English fallback pattern used elsewhere in the file.

VascoSch92: I see... Actually I had the same problem in another PR :-) 

Thanks to solve the conflicts. 

Waiting for the tests and then I think we are good to go.

muff-c:
Yep, ready now. Repo checks are green, including the snapshot rerun after the `update-snapshots` label. Vercel still needs team authorization on your side.

VascoSch92 (approved):
@muff-c Thanks :-)

PR #1298, opened 2026-06-10
bump automation to 1.0.0a7

tofarr:
HUMAN:

This is just a dependency bump

AGENT:

Verified with `npm run make-i18n && node_modules/.bin/vitest run __tests__/scripts/check-sdk-version-sync.test.ts __tests__/scripts/docs-version-sync.test.ts`:

```
 ✓ __tests__/scripts/docs-version-sync.test.ts (4 tests) 4ms
 ✓ __tests__/scripts/check-sdk-version-sync.test.ts (19 tests) 51ms

 Test Files  2 passed (2)
      Tests  23 passed (23)
```

Also confirmed `npm run typecheck` exits 0.

---

## Why

`openhands-automation` `1.0.0a7` is now available on PyPI and pins `openhands-sdk==1.27.0`, matching the `agentServer` version already in `config/defaults.json`. The previous `automationSdk` value (`1.22.1`) was stale.

## Summary

- Bump `versions.automation` from `1.0.0a6` → `1.0.0a7` in `config/defaults.json`
- Bump `versions.automationSdk` from `1.22.1` → `1.27.0` (matches the SDK dependency of `1.0.0a7`)
- Add `config/defaults.json` to the path triggers of `.github/workflows/sdk-version-sync.yml` so the PyPI metadata check also fires when the config file is updated

## Issue Number

N/A

## How to Test

To verify the automation version is real:
```bash
curl -s "https://pypi.org/pypi/openhands-automation/1.0.0a7/json" | \
  python3 -c "import json,sys; d=json.load(sys.stdin); print([r for r in d['info']['requires_dist'] if 'openhands' in r])"
# Expected: ['openhands-sdk==1.27.0', 'openhands-workspace==1.27.0']
```

## Video/Screenshots

image

image

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [x] Docs / chore


malhotra5 (changes requested):
I think we have a CI that performs a version compatibility check already: https://github.com/OpenHands/agent-canvas/blob/main/.github/workflows/sdk-version-sync.yml

The reason I'm hesitant about this particular implementation is because it requires manually specifying the automation sdk version. This is brittle as we can bump the default defined here without actually bumping the sdk version in the automation BE as a means to get past the CI

tofarr:
Addressed in be6758c7 — the static `versions.automationSdk === versions.agentServer` test has been removed. The `sdk-version-sync` workflow already covers the meaningful invariant (PyPI-level consistency between the released `openhands-automation` package and `versions.automationSdk`). The static test would have enforced equality at all times, but the script explicitly allows `automationSdk` to lag `agentServer` while an automation release is pending — making it both unnecessary and incorrect. The PR is now a pure version bump plus the `config/defaults.json` workflow path trigger.

malhotra5 (approved):
LGTM

config/defaults.json
```diff
@@ -4,8 +4,8 @@
   "versions": {
     "agentServer": "1.27.0",
     "agentCanvas": "1.0.0-rc.3",
-    "automation": "1.0.0a6",
-    "automationSdk": "1.22.1",
+    "automation": "1.0.0a7",
+    "automationSdk": "1.27.0",
```
malhotra5, line 8:
```suggestion
```
I'm not sure why we have this specification, we should remove it if possible

tofarr: Not a hallucination — `automationSdk` is actually the core input to the `sdk-version-sync` workflow you pointed to in your first review comment. The script (`scripts/check-sdk-version-sync.mjs`) reads `versions.automationSdk` as the expected SDK version, then fetches the released `openhands-automation` package from PyPI and verifies that its actual SDK dependencies match. It requires the field and exits 1 if it is missing — so removing it would break that workflow entirely.

The reason it exists as a separate field from `versions.agentServer` is exactly the allowance you noted earlier: the two can legitimately differ while an `openhands-automation` release is pending. `automationSdk` records what the *currently pinned* automation release depends on, which may lag `agentServer` by a version or two during that window.

PR #1370, opened 2026-06-15
Fix stale agent context datetime

DevinVinson:
HUMAN:

AGENT:

---

## Summary
Fix a bug where new conversations could inherit a stale `agent_context.current_datetime` from persisted settings instead of getting a fresh runtime timestamp.

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


DevinVinson:
Discussed in slack and an sdk side change will be a better fix. Closing this out.

PR #1439, opened 2026-06-21
docs: align sandbox quickstart defaults

koriyoshi2041:
HUMAN:

AGENT:

Docs-only change. I checked the changed markdown for trailing whitespace and searched the quickstart docs to confirm the old defaults were removed.

---

## Why

A reviewer on #1428 suggested a follow-up PR to avoid creating both `projects` and `Projects` host folders, and to avoid a fixed old Docker image tag in the quickstart.

## Summary

- Use `$HOME/Projects` as the host-side default for the Docker sandbox project directory in Unix and Windows quickstarts.
- Use `ghcr.io/openhands/agent-canvas:latest` instead of the old fixed `1.0.0-rc.11` image tag.
- Keep the container mount path as `/projects`.

## Issue Number

Follow-up from review discussion in #1428.

## How to Test

- `git diff --check`
- `rg -n '\$HOME/projects|Join-Path \$HOME "projects"|agent-canvas:1\.0\.0-rc\.11' README.md README.windows.md || true`

## Video/Screenshots

Not applicable; docs-only quickstart change.

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [x] Docs / chore

## Notes

Risk is low. This only changes documented host defaults and image tag; it does not change the container's internal `/projects` path or runtime code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,14 +90,14 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 **macOS / Linux:**
 
 ```sh
-export PROJECTS_PATH="$HOME/Projects"  # directory containing your project folders
+export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
 
 docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:latest
+  ghcr.io/openhands/agent-canvas:1.0.0-rc.11
 ```
 
 **Windows (PowerShell / Windows Terminal):** See [README.windows.md](./README.windows.md) for the equivalent commands.
--- a/README.md
+++ b/README.md
@@ -90,14 +90,14 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 **macOS / Linux:**
 
 ```sh
-export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
+export PROJECTS_PATH="$HOME/Projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
 
 docker run -it --rm \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,14 +90,14 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
...
-export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
+export PROJECTS_PATH="$HOME/Projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
 
 docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:1.0.0-rc.11
+  ghcr.io/openhands/agent-canvas:latest
```
VascoSch92: This should not be changed. 

In fact, this will be reverted before next release anyway. See Release.md file.

README.md
```diff
@@ -90,14 +90,14 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 **macOS / Linux:**
 
 ```sh
-export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
+export PROJECTS_PATH="$HOME/Projects"  # directory containing your project folders
```
VascoSch92: Why this capitalisation?

VascoSch92 (changes requested):

koriyoshi2041:
Thanks for the review. Since the PR only changed the defaults you flagged, I restored the existing lowercase projects path and release tag and am closing this rather than leaving a no-op docs PR open.

README.md
mysteryx93: Because Linux added Projects as a standard folder, alongside Documents, Downloads, etc. So you end up with both 'Projects' and 'projects' folders. In other words, the capitalization is already there on the host. The question is whether you want to use the system folder, or have your own separate folder. If you want to have a separate folder, then it should be intentional and hold a different name that doesn't cause confusion.

PR #1457, opened 2026-06-23
fix: pickFallbackBackend ignores backend health and kind

### LINKED ISSUE #1174: [Bug]: pickFallbackBackend ignores backend health & kind — Settings → LLMs shows "No backend is configured" even when a healthy local backend is registered
### Summary

When there's no valid backend selection (fresh start, or the selected backend was removed), the active backend is chosen by `pickFallbackBackend`, which returns `backends[0]` — the first registered backend in insertion order, ignoring both its `kind` and its recorded health. If index-0 is a cloud backend or a dead local backend, `getEffectiveLocalBackend()` resolves to `null`/dead, and every local-protocol call (verified models, LLM profiles, conversation CRUD) throws `No backend is configured.` — even when a healthy local backend is registered further down the list. A synchronous health store already exists at the point of selection but isn't consulted.

Split out of #1093 (its Docker/local part was handled in #1081).

### Environment

macOS · Docker (`ghcr.io/openhands/agent-canvas`) · 1.0.0-beta.7 · verified against `main` @ `df8a573`

### Steps to Reproduce

1. Register two backends so that index-0 is either a cloud backend or a now-stopped local backend, and a healthy local backend is *not* index-0.
2. Have no valid selection (fresh load, or delete the currently-selected backend).
3. Open **Settings → LLMs**.

### Actual Behavior

`No backend is configured.` toast; empty LLM Provider dropdown; no LLM profiles. The healthy local backend is never tried.

### Expected Behavior

Fallback selection resolves to a healthy local backend when one is registered; verified models and profiles load.

### Root Cause

`pickFallbackBackend` (`src/api/backend-registry/active-store.ts:36-38`):

```ts
function pickFallbackBackend(backends: Backend[]): Backend {
  return backends[0] ?? NO_BACKEND;
}
```

Used by `computeSnapshot` when no selection resolves (`active-store.ts:58-62`). The chosen backend flows to `getEffectiveLocalBackend()` (`:94-98`), which returns `null` if it isn't a live local backend; local-protocol calls then throw `NoBackendAvailableError("No backend is configured.")` (`agent-server-client-options.ts:22-27`, `:55-57`). Selection consults neither `kind` nor the health store.

### Proposed Fix

A synchronous health store already exists at this exact point (`health-store.ts:29` `getBackendHealthEntry`; fed by the 10s probe in `use-backends-health.ts`; `disabled` persists in localStorage). Rank candidates by recorded health (and prefer a local backend, since that's what local-protocol calls need) instead of taking index-0 (illustrative, untested):

```ts
import { getBackendHealthEntry } from "./health-store";

function failScore(b: Backend): number {
  const h = getBackendHealthEntry(b.id);
  if (!h) return 0;            // no recorded failures → best
  if (h.disabled) return 1000; // probes gave up → worst
  return h.consecutiveFailures;
}

function pickFallbackBackend(backends: Backend[]): Backend {
  if (backends.length === 0) return NO_BACKEND;
  const ranked = [...backends].sort((a, b) => failScore(a) - failScore(b)); // stable: index-0 stays the tiebreak
  return ranked[0] ?? NO_BACKEND;
}
```

### Caveats / open questions

1. **Cold-start ambiguity** — `null` health can't distinguish "healthy" from "never probed"; before the first probe a stale-but-unflagged backend can still win. localStorage persistence means it self-corrects after the first session.
2. **Should the fallback prefer a *local* backend over a cloud one?** Local-protocol calls need a local backend, but the current design may intentionally allow a cloud fallback — needs a product call.
3. **Explicit cloud selection** still yields `getEffectiveLocalBackend() === null` by design (docstring in `active-store.ts:86-93`). That's the literal #1093 repro but appears intentional — if verified models/profiles should work on a cloud backend, that's a *separate* issue.

### Related

#1093 (parent) · #1081 (Docker/local fix)

mvanhorn:
HUMAN:

AGENT:

This change is library-internal to the backend registry (`src/api/backend-registry/active-store.ts`). I verified it with the package's own vitest suite against the touched and adjacent modules; there is no standalone way to exercise this resolver end-to-end without a multi-backend agent-server setup, so the verification is the unit suite plus the reproduction notes under How to Test below.

---

## Why

When there is no valid active backend selection (a fresh start, or the previously selected backend was removed), `pickFallbackBackend()` returned `backends[0]` — the first registered backend in insertion order — ignoring both its `kind` and its recorded health.

If index 0 is a cloud backend, or a local backend that has already been marked `disabled` in the persisted health store, `getEffectiveLocalBackend()` resolves to null. Every local-protocol call (verified models, LLM profiles, conversation CRUD) then throws "No backend is configured" — the symptom reported in the issue: Settings -> LLMs shows "No backend is configured" even when a healthy local backend is registered further down the list. A synchronous health store (`health-store.ts`) already exists at this selection point but was not consulted.

## Summary

- `pickFallbackBackend()` now consults the synchronous, persisted health store (`getBackendHealthEntry`) when resolving the fallback, and prefers a healthy local backend over a cloud backend or a backend already marked unhealthy.
- Selection priority on the fallback path: (1) a `local` backend not marked `disabled`, then (2) any `local` backend (so the effective-local resolver still resolves when all locals are unhealthy), then (3) the prior deterministic `backends[0] ?? NO_BACKEND` for the registry-has-no-local case.
- Scope is limited to the fallback path: explicit selections, including cloud, are untouched. No new exported API surface; the health store is read-only at this call site.

## Issue Number

Fixes #1174

## How to Test

```
npm install
npx vitest run __tests__/api/backend-registry/active-store.test.ts
```

All tests pass. New cases cover: a cloud backend registered first with a healthy local later (picks the local), an unhealthy local backend skipped for a healthy one further down, all-local-unhealthy still resolving to a local backend (no crash, deterministic), a single local backend at index 0, the registry-has-no-local case still returning `backends[0]`, and an explicit cloud selection remaining untouched.

Manual reproduction of the original bug: register a cloud backend (or a local backend whose health has tripped the failure cap) at the first registry position and a healthy local backend after it, with no explicit active selection. Before this change, Settings -> LLMs reports "No backend is configured"; after it, the healthy local backend is selected.

## Video/Screenshots

Not applicable — this is an internal resolver change with no UI surface of its own; evidence is the unit suite above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix reads health lazily at fallback-resolution time and intentionally does not subscribe `active-store` to live `health-store` changes — re-selecting a different backend mid-session would also need the provider's bootstrap/cache-invalidation flow, which is out of scope for this fix. The resolver picks up the correct backend on the next registry/selection write or reload, which is the path the reported symptom occurs on.

### DIFF
--- a/__tests__/api/backend-registry/active-store.test.ts
+++ b/__tests__/api/backend-registry/active-store.test.ts
@@ -9,11 +9,17 @@ import {
   subscribeActiveBackend,
 } from "#/api/backend-registry/active-store";
 import { SEEDED_DEFAULT_BACKEND_ID } from "#/api/backend-registry/default-backend";
+import { MAX_CONSECUTIVE_FAILURES } from "#/api/backend-registry/health-storage";
+import {
+  __resetHealthStoreForTests,
+  recordBackendFailure,
+} from "#/api/backend-registry/health-store";
 import type { Backend } from "#/api/backend-registry/types";
 
 beforeEach(() => {
   window.localStorage.clear();
   window.sessionStorage.clear();
+  __resetHealthStoreForTests();
   __resetActiveStoreForTests();
 });
 
@@ -21,6 +27,7 @@ afterEach(() => {
   window.localStorage.clear();
   window.sessionStorage.clear();
   vi.unstubAllEnvs();
+  __resetHealthStoreForTests();
   __resetActiveStoreForTests();
 });
 
@@ -39,6 +46,20 @@ const localBackend: Backend = {
   apiKey: "k",
   kind: "local",
 };
+
+const secondLocalBackend: Backend = {
+  id: "local-2",
+  name: "Local 2",
+  host: "http://localhost:9001",
+  apiKey: "k2",
+  kind: "local",
+};
+
+function markBackendUnhealthy(id: string): void {
+  for (let i = 0; i < MAX_CONSECUTIVE_FAILURES; i += 1) {
+    recordBackendFailure(id, new Error("connection failed"));
+  }
+}
 
 describe("active-store", () => {
   it("uses the no-backend sentinel when no backend details are available", () => {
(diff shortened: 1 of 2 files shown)


jamiechicago312:
Hey @mvanhorn 

Thank you for submitting this PR. I will ask maintainer in [Slack's](go.openhands.dev/slack) #proj-agent-canvas channel to review.

Can you fix the failing test "Pr Description Check / Validate PR description"? It required you to edit the description, and checkmark the tested by a human.

Thank you!

jamiechicago312:
> Hey @mvanhorn
> 
> Thank you for submitting this PR. I will ask maintainer in [Slack's](go.openhands.dev/slack) #proj-agent-canvas channel to review.
> 
> Can you fix the failing test "Pr Description Check / Validate PR description"? It required you to edit the description, and checkmark the tested by a human.
> 
> Thank you!

@ me when you've got it passing all test minus vercel and then i can request a review tysm

neubig:

PR #1474, opened 2026-06-24
feat: stream OpenHands agent responses

### LINKED ISSUE #1471: Support Chat streaming for OpenHands agent
Right now the ACP agents can stream text during conversations but the OpenHands agent does not. It would be great if the default experience users expect for agent conversations was consistent in Agent Canvas.

VascoSch92:
HUMAN:

Add streaming for OH agent.

AGENT:

End-to-end verification was done against a real local stack, not just unit tests.

**Stack:** `npm run dev` (ingress `http://localhost:8000`, Vite frontend, OpenHands Agent Server `openhands-sdk v1.28.1` from PyPI), real LLM profile `openhands/gpt-5.5`. The UI was driven with Playwright (create conversation, dismiss onboarding/consent, send messages) while capturing the conversation websocket frames.

**1. Streaming works at conversation start.** Sending "Say hello in one short sentence." produced `StreamingDeltaEvent` frames token-by-token over `/sockets/events/<id>`, e.g.:
```
StreamingDeltaEvent content="<think>The"
StreamingDeltaEvent content=" user is"
... (tokens) ...
StreamingDeltaEvent content="Hello! I'm"
MessageEvent source=agent text="<think>…</think>…Hello! I'm here to help…"
```
The streamed deltas reconcile to a single bubble (existing `finalizeStreamingDeltasInPlace`); the duplicate `MessageEvent` seen on the wire shares the same `id` and is deduped by the store.

**2. A/B confirming the root cause of the "message shown twice" report.** Same prompt, `stream` toggled:
- `stream` omitted/false → bubble: `Hello! How can I help you today?` (litellm separates reasoning when not streaming).
- `stream: true` without the fix → bubble: `The user is asking me to say hello… [reasoning] … Hello! I'm here to help…` (the model emits inline `<think>` reasoning, which leaked into the message text).
- With the `splitInlineThink` fix → bubble: `Hello! I'm here to help you with any tasks you need.` and the reasoning is rendered in the collapsed **Thinking** section. Verified identical after a page reload (persisted message path).

**3. Streaming survives a mid-conversation model switch.** Initial message streamed (2 deltas) → switched model with the `/model gpt-5.5` command (routes through `switchProfile` → `POST /api/conversations/{id}/switch_llm`) → next message still streamed (3 deltas). Before the fix this dropped to 0 deltas because the switch payload defaulted to `stream=false`.

**CI checks run locally** (after syncing `node_modules` to the lockfile):
- `npm run typecheck` → 0 errors
- `npm run build` → success
- `npx vitest run` on affected suites → 298 passed (incl. new `splitInlineThink` tests and updated start/switch payload tests)
- pre-commit hook (eslint --fix, prettier, staged tsc, translation completeness) passed on commit

---

## Why

The ACP agents stream their responses token-by-token, but the OpenHands agent did not — it rendered each reply only once it was complete. This is an inconsistent conversation experience across agent types (issue #1471).

While enabling streaming, two correctness issues surfaced and are fixed here:
- Models that emit chain-of-thought as an inline `<think>…</think>` block (e.g. `openhands/gpt-5.5`) had that reasoning leak into the visible message — it reads like the message is shown twice.
- Switching the model mid-conversation silently turned streaming back off.

## Summary

- Enable LLM token streaming for the OpenHands (non-ACP) agent by setting `stream: true` on the LLM config — in both the conversation-start payload (`buildConfiguredOpenHandsAgentSettings`) and the mid-conversation `switch_llm` payload (`switchProfile`). The agent-server only emits `StreamingDeltaEvent`s when an LLM has `stream=True`; the frontend already renders them.
- Add `splitInlineThink` and route inline `<think>…</think>` reasoning into the existing collapsible "Thinking" section instead of the message bubble — applied to both the streamed-delta renderer and the persisted agent-message renderer so live and reloaded views match. Safe no-op when there is no `<think>` block.

## Issue Number

Closes #1471

## How to Test

1. `npm install && npm run dev`, open `http://localhost:8000/`.
2. Configure an LLM (a reasoning model such as `openhands/gpt-5.5` best demonstrates the `<think>` handling).
3. Start a New Chat with the **OpenHands** agent and send a message — the reply should stream in token-by-token (live typing), with any model reasoning collapsed under **Thinking** and only the answer in the bubble.
4. Mid-conversation, switch the model (model pill in the chat input, or the `/model <profile>` command) and send another message — it should still stream.
5. Reload the page — the message stays clean (no `<think>` text in the bubble).

## Video/Screenshots

https://github.com/user-attachments/assets/a7f93ce7-e658-4eaa-906d-c270d1887c03

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- This is the display-side fix for inline reasoning: the frontend parses `<think>` out at render time, but the block is still written into the **persisted** message content by the backend when streaming. Making the saved message clean too would be a `software-agent-sdk` change (separating reasoning during streaming) — out of scope here.
- ACP agents are unaffected: they stream unconditionally on the backend and their config/switch paths build no `llm` object.

---

| Component | Value |
|---|---|
| **Commit** | `77d4aed9e3f7f177f81f8ce47326d9b5330a9ec6` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-77d4aed
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-77d4aed
```

```
ghcr.io/openhands/agent-canvas:sha-77d4aed-amd64
ghcr.io/openhands/agent-canvas:vasco-streaming-amd64
ghcr.io/openhands/agent-canvas:pr-1474-amd64
ghcr.io/openhands/agent-canvas:sha-77d4aed-arm64
ghcr.io/openhands/agent-canvas:vasco-streaming-arm64
ghcr.io/openhands/agent-canvas:pr-1474-arm64
ghcr.io/openhands/agent-canvas:sha-77d4aed
ghcr.io/openhands/agent-canvas:vasco-streaming
ghcr.io/openhands/agent-canvas:pr-1474
```

- Each tag (e.g., `sha-77d4aed`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-77d4aed-amd64`) are also available if needed

### DIFF
--- a/src/components/conversation-events/chat/event-thought-helpers.ts
+++ b/src/components/conversation-events/chat/event-thought-helpers.ts
@@ -44,57 +44,41 @@ export const hasNonEmptyThought = (action: ActionEvent): boolean =>
   getActionThoughtText(action).trim().length > 0;
 
 /**
- * Splits an inline `<think>…</think>` reasoning block out of assistant
- * message content.
+ * Splits a leading `<think>…</think>` reasoning block out of assistant content
+ * so it renders in the collapsible thinking section, not the message bubble.
+ * Some models stream reasoning inline instead of via `reasoning_content`.
  *
- * Some models stream their chain-of-thought as an inline `<think>` block
- * inside the regular `content` instead of the dedicated `reasoning_content`
- * field. litellm extracts that block when a completion is NOT streamed, but
- * raw streamed deltas keep it inline — so a streamed OpenHands message would
- * otherwise render the reasoning as visible message text. This routes the
- * reasoning to the same collapsible "thinking" section used for
- * `reasoning_content`.
- *
- * Returns the extracted `reasoning` (joined across blocks) and the remaining
- * `message`. When there is no `<think>` block the content is returned
- * unchanged as `message`, so this is a safe no-op for normal messages. An
- * unclosed `<think>` (mid-stream) is treated as reasoning in full, so partial
- * thinking never leaks into the message bubble.
+ * Conservative to avoid mangling normal messages: only a `<think>` at the very
+ * start is touched (later occurrences, e.g. quoted in docs, stay verbatim),
+ * only the first block is peeled, and an unclosed leading `<think>` is reasoning
+ * only while `streaming` — in a finalized message it's literal output.
  */
 export const splitInlineThink = (
   content: string,
+  options?: { streaming?: boolean },
 ): { reasoning: string; message: string } => {
   const OPEN = "<think>";
   const CLOSE = "</think>";
-  if (!content.includes(OPEN)) {
+
+  // Only a <think> at the very start is reasoning.
+  const leading = content.replace(/^\s+/, "");
+  if (!leading.startsWith(OPEN)) {
     return { reasoning: "", message: content };
   }
 
-  const reasoning: string[] = [];
-  let message = "";
-  let rest = content;
+  const afterOpen = leading.slice(OPEN.length);
+  const close = afterOpen.indexOf(CLOSE);
 
-  while (rest.length > 0) {
-    const open = rest.indexOf(OPEN);
-    if (open === -1) {
-      message += rest;
-      break;
-    }
-    message += rest.slice(0, open);
-    const afterOpen = rest.slice(open + OPEN.length);
-    const close = afterOpen.indexOf(CLOSE);
-    if (close === -1) {
-      // Unclosed block: still streaming — everything left is reasoning.
-      reasoning.push(afterOpen);
-      break;
-    }
-    reasoning.push(afterOpen.slice(0, close));
-    rest = afterOpen.slice(close + CLOSE.length);
+  if (close === -1) {
+    // Unclosed: reasoning-in-progress while streaming, else literal output.
+    return options?.streaming
+      ? { reasoning: afterOpen.trim(), message: "" }
+      : { reasoning: "", message: content };
   }
 
   return {
-    reasoning: reasoning.join("\n\n").trim(),
-    message: message.trim(),
+    reasoning: afterOpen.slice(0, close).trim(),
+    message: afterOpen.slice(close + CLOSE.length).trim(),
   };
 };
 
--- a/src/components/conversation-events/chat/event-thought-helpers.ts
+++ b/src/components/conversation-events/chat/event-thought-helpers.ts
@@ -44,6 +44,61 @@ export const hasNonEmptyThought = (action: ActionEvent): boolean =>
   getActionThoughtText(action).trim().length > 0;
 
 /**
+ * Splits an inline `<think>…</think>` reasoning block out of assistant
+ * message content.
+ *
+ * Some models stream their chain-of-thought as an inline `<think>` block
+ * inside the regular `content` instead of the dedicated `reasoning_content`
+ * field. litellm extracts that block when a completion is NOT streamed, but
+ * raw streamed deltas keep it inline — so a streamed OpenHands message would
+ * otherwise render the reasoning as visible message text. This routes the
+ * reasoning to the same collapsible "thinking" section used for
+ * `reasoning_content`.
+ *
+ * Returns the extracted `reasoning` (joined across blocks) and the remaining
+ * `message`. When there is no `<think>` block the content is returned
+ * unchanged as `message`, so this is a safe no-op for normal messages. An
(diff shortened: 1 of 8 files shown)


src/components/conversation-events/chat/event-thought-helpers.ts
```diff
@@ -43,6 +43,61 @@ export const getReasoningContent = (action: ActionEvent): string => {
...
+ * Splits an inline `<think>…</think>` reasoning block out of assistant
+ * message content.
+ *
+ * Some models stream their chain-of-thought as an inline `<think>` block
+ * inside the regular `content` instead of the dedicated `reasoning_content`
+ * field. litellm extracts that block when a completion is NOT streamed, but
+ * raw streamed deltas keep it inline — so a streamed OpenHands message would
+ * otherwise render the reasoning as visible message text. This routes the
+ * reasoning to the same collapsible "thinking" section used for
+ * `reasoning_content`.
```
enyst: This is a bit of chaos out there, unfinished tags, etc, I think maybe the way this PR is doing it is the best we can do…

enyst (approved):
Thank you! This looks great ❤️

src/components/conversation-events/chat/event-thought-helpers.ts
VascoSch92: This is because minimax give back reasoning as `<thinking> bla bla bla <thinking>`

src/components/conversation-events/chat/event-thought-helpers.ts
```diff
@@ -43,6 +43,61 @@ export const getReasoningContent = (action: ActionEvent): string => {
...
+ *
+ * Returns the extracted `reasoning` (joined across blocks) and the remaining
+ * `message`. When there is no `<think>` block the content is returned
+ * unchanged as `message`, so this is a safe no-op for normal messages. An
+ * unclosed `<think>` (mid-stream) is treated as reasoning in full, so partial
+ * thinking never leaks into the message bubble.
+ */
+export const splitInlineThink = (
```
hieptl, line 56: **question (blocking):**

`splitInlineThink` currently corrupts legitimate messages that contain the literal `<think>` substring.

The function is applied to every persisted agent message (`user-assistant-event-message.tsx:26-29`) as well as every streamed delta, and it matches any occurrence of `<think>` in the content. As a result, agent responses that simply reference these tags—for example in HTML/XML examples, code snippets, or prompt-engineering documentation—can be unintentionally modified.

For example, the message:

> "You can wrap reasoning in `<think>` and `</think>` tags."

is rendered as:

> "You can wrap reasoning in `` tags."

with the literal tags removed and the text between them incorrectly categorized as "Thinking".

The behavior is even more problematic when only an opening `<think>` tag is present. In that case, the unclosed-tag branch moves the entire remainder of the message into the collapsed Thinking section. Prior to this PR, such content rendered verbatim, so this appears to be a regression for the non-trivial case where an agent intentionally emits the literal string `<think>`.

Would it make sense to narrow the matching logic to the specific pattern produced by LiteLLM's non-streamed extraction—for example, a `<think>` block anchored at the beginning of the content—rather than matching any occurrence? Alternatively, we could limit the split behavior to streamed messages only. Either approach would preserve the intended fix while avoiding corruption of otherwise valid message content.

Thank you! 🙏

hieptl (changes requested):

src/components/conversation-events/chat/event-thought-helpers.ts
enyst: Yes, there might be no ideal solution here. Litellm is doing the same thing or was doing the same thing: stripping `<think> </think>` and try to add it as reasoning. That has some false positives occasionally.

The trouble is that some LLMs are doing it anyway and LiteLLM doesn’t extract them. 😥

We could narrow it to only streaming?

VascoSch92: Actually, there is another problem which has nothing to do with streaming. if you try to write somethin like <word> in agent canvas: this would be deleted :-) This is because of a sanitize that we do.

VascoSch92:
Hey @hieptl 

thanks for the comment. 

It should be fixed. 

However, it is a little difficult to test because of this: https://github.com/OpenHands/agent-canvas/pull/1481

hieptl (approved):
Thank you! 🙏

PR #1479, opened 2026-06-24
fix: redact secrets in automation run logs

### LINKED ISSUE #1470: Security: SESSION_API_KEY leaked in automation run command logs
## Summary

`SESSION_API_KEY` (and potentially other secrets) are being exposed in plain text in automation run command logs.

## Observed behavior

When an automation run times out (or potentially in other log paths), the full shell command is recorded in the log output, including the `export SESSION_API_KEY=...` assignment verbatim. Example from a timed-out automation run:

```
Command timed out after 300 seconds: mkdir -p /root/.openhands/agent-canvas/workspaces/automation-runs/27fa5d48-a741-416d-a5c3-109d1d336a51 \
  && tar xzf /tmp/automation-27fa5d48-a741-416d-a5c3-109d1d336a51.tar.gz \
       -C /root/.openhands/agent-canvas/workspaces/automation-runs/27fa5d48-a741-416d-a5c3-109d1d336a51 \
  && rm -f /tmp/automation-27fa5d48-a741-416d-a5c3-109d1d336a51.tar.gz \
  && cd /root/.openhands/agent-canvas/workspaces/automation-runs/27fa5d48-a741-416d-a5c3-109d1d336a51 \
  && export AGENT_SERVER_URL='http://localhost:18000' \
  && export SESSION_API_KEY=<REDACTED>
```

The `SESSION_API_KEY` value appeared in full in the actual log (redacted here).

## Expected behavior

Sensitive environment variables (`SESSION_API_KEY`, `AUTOMATION_LOCAL_API_KEY`, and any other credential-bearing values) must **never** appear in log output, error messages, exception traces, or timeout messages.

## Impact

- Anyone with access to the automation run logs (e.g. log aggregators, CI artifacts, monitoring dashboards) can read the session API key.
- The key grants full access to the agent-server API for the duration it remains valid.

## Suggested fixes

1. **Mask before logging**: When the automation backend constructs the command string for execution, redact any value assigned to a known-sensitive variable (e.g. any `export *KEY*=`, `export *SECRET*=`, `export *TOKEN*=`) before writing it to logs.
2. **Pass secrets via environment, not inline args**: Instead of embedding `export SESSION_API_KEY=<value>` in the shell command string, pass secrets through the subprocess `env` parameter (Python `subprocess.run(..., env={...})`) so the value never appears in the command string at all.
3. **Scrub timeout/error messages**: Ensure any exception handler or timeout handler that logs the raw command applies the same masking.
4. **Audit other log sites**: Review all places where automation commands are logged (startup, retry, success, failure) to confirm no other credential-bearing values are emitted.

## References

- AGENTS.md: _"Failure messages around live Agent Server settings must not print response bodies from credential-bearing requests."_
- `scripts/dev-with-automation.mjs` – `AUTOMATION_LOCAL_API_KEY` and `SESSION_API_KEY` generation

---

nehaprasad-dev:
#### Summary
- Redact sensitive export values in automation run logs before they show in the UI.

#### Test plan
- npm test -- __tests__/utils/redact-shell-secrets.test.ts __tests__/components/automations/detail/run-logs-modal.test.tsx

#### Video/Screenshots
- N/A

#### Type
- Bug fix

fixes : #1470 (UI only)

### DIFF
--- a/src/utils/redact-shell-secrets.ts
+++ b/src/utils/redact-shell-secrets.ts
@@ -1,6 +1,5 @@
 const REDACTED = "<redacted>";
 
-// Mirrors openhands-sdk SECRET_KEY_PATTERNS (substring match, case-insensitive).
 const SECRET_KEY_PATTERNS = [
   "AUTHORIZATION",
   "COOKIE",
@@ -17,7 +16,6 @@ export function isSensitiveEnvVarName(name: string): boolean {
   return SECRET_KEY_PATTERNS.some((pattern) => upper.includes(pattern));
 }
 
-// export VAR='value', export VAR="value", or export VAR=unquoted
 const EXPORT_ASSIGNMENT =
   /export\s+([A-Za-z_][A-Za-z0-9_]*)\s*=\s*('(?:\\'|[^'])*'|"(?:\\"|[^"])*"|[^\s&;|\\]+)/g;
 
@@ -31,11 +29,6 @@ function redactAssignment(
   return `${prefix}'${REDACTED}'`;
 }
 
-/**
- * Redact credential-bearing shell environment assignments before showing
- * automation run logs. Covers inline `export VAR=value` chains produced by
- * the automation backend when bootstrapping a run workspace.
- */
 export function redactShellSecrets(text: string): string {
   if (!text) return text;
 
--- a/src/utils/redact-shell-secrets.ts
+++ b/src/utils/redact-shell-secrets.ts
@@ -1,0 +1,43 @@
+const REDACTED = "<redacted>";
+
+// Mirrors openhands-sdk SECRET_KEY_PATTERNS (substring match, case-insensitive).
+const SECRET_KEY_PATTERNS = [
+  "AUTHORIZATION",
+  "COOKIE",
+  "CREDENTIAL",
+  "KEY",
+  "PASSWORD",
+  "SECRET",
+  "SESSION",
+  "TOKEN",
+] as const;
+
+export function isSensitiveEnvVarName(name: string): boolean {
+  const upper = name.toUpperCase();
+  return SECRET_KEY_PATTERNS.some((pattern) => upper.includes(pattern));
+}
+
+// export VAR='value', export VAR="value", or export VAR=unquoted
+const EXPORT_ASSIGNMENT =
+  /export\s+([A-Za-z_][A-Za-z0-9_]*)\s*=\s*('(?:\\'|[^'])*'|"(?:\\"|[^"])*"|[^\s&;|\\]+)/g;
+
+function redactAssignment(
+  match: string,
+  name: string,
+  valuePart: string,
+): string {
+  if (!isSensitiveEnvVarName(name)) return match;
+  const prefix = match.slice(0, match.length - valuePart.length);
+  return `${prefix}'${REDACTED}'`;
+}
+
+/**
+ * Redact credential-bearing shell environment assignments before showing
+ * automation run logs. Covers inline `export VAR=value` chains produced by
+ * the automation backend when bootstrapping a run workspace.
+ */
+export function redactShellSecrets(text: string): string {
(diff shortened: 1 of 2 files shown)


src/utils/redact-shell-secrets.ts
hieptl, line 1: **suggestion (blocking):**

Secret content may be partially exposed when a value contains an unescaped single quote.

The single-quote matching branch, `'(?:\\'|[^'])*'`, terminates at the first literal `'`. Because shell single-quoted strings cannot directly contain a single quote, a secret such as `pa'ss` is represented as `'pa'\''ss'`. In this case, the regex matches only `'pa'`, redacts that portion as `<redacted>`, and leaves `\''ss'` visible in the log, resulting in partial disclosure of the secret (`ss`).

A similar partial-leak scenario can occur with unterminated quoted values. For example, with `export TOKEN="secret with no close`, the unquoted fallback pattern `[^\s&;|\\]+` matches only `"secret`, while the remainder of the value remains visible in the log.

Thank you! 🙏

src/utils/redact-shell-secrets.ts
```diff
@@ -0,0 +1,36 @@
...
+  "SECRET",
+  "SESSION",
+  "TOKEN",
+] as const;
+
+export function isSensitiveEnvVarName(name: string): boolean {
+  const upper = name.toUpperCase();
+  return SECRET_KEY_PATTERNS.some((pattern) => upper.includes(pattern));
+}
+
```
hieptl, line 18: Redaction currently applies only to the `export VAR=...` pattern, while several common mechanisms for exposing secrets remain unaddressed.

Examples include inline environment variable assignments (`SESSION_TOKEN=abc some-cmd`), standalone assignments (`API_KEY=abc` on its own line), `env VAR=...` invocations, and any values that are subsequently echoed or printed by commands (for example, `echo $TOKEN` or `env` output). These cases pass through without redaction.

The PR description explicitly scopes the implementation to redacting values in `export` statements, so this behavior may be intentional and acceptable for the current scope. However, given that the function is named `redactShellSecrets` and is intended to prevent secrets from being displayed, the narrower coverage may warrant additional documentation, clarification, or a follow-up enhancement. These patterns are commonly encountered in shell logs and could reasonably be expected to fall within the function's remit.

Thank you! 🙏

src/utils/redact-shell-secrets.ts
```diff
@@ -0,0 +1,36 @@
...
+const SECRET_KEY_PATTERNS = [
+  "AUTHORIZATION",
+  "COOKIE",
+  "CREDENTIAL",
+  "KEY",
+  "PASSWORD",
+  "SECRET",
+  "SESSION",
+  "TOKEN",
```
hieptl, line 11: **suggestion (blocking):**

The current substring-based key matching approach may result in over-redaction.

For example, `upper.includes("KEY")` will match identifiers such as `MONKEY`, `KEYBOARD`, and `KEYWORD`, causing them to be redacted even when they do not represent sensitive values. This behavior is conservative and favors over-redaction rather than risking exposure of secrets, so the impact is primarily on log readability rather than security.

At the same time, some commonly used secret-related identifiers, such as `JWT`, `PAT`, and `NONCE`, are not currently included in the matching logic. While this does not necessarily introduce an immediate issue, it highlights that the current heuristic may both over-match certain benign identifiers and miss other token-related naming conventions.

Thank you! 🙏

hieptl (changes requested):
Hello @nehaprasad-dev ,

Thank you for creating this pull request!

I've left a few comments. Could you please add tests to cover the changes introduced in this PR? Additionally, it would be helpful to include a demo screenshot demonstrating the resulting output.

Also, could you please update the pull request description to follow our PR template?

Thank you very much for your contribution. I appreciate your time and effort in working on this! 🙏

neubig:

PR #1487, opened 2026-06-25
fix: retry transient backend disconnects before showing backend picker

### LINKED ISSUE #1420: Retry transient backend disconnects before showing backend picker
## Problem

Agent Canvas can move users to the backend recovery / picker UI after a single transient backend failure during startup. This is disruptive when the agent server or tunnel is only briefly slow.

Two paths are currently too sensitive:

- The initial config query calls `/server_info`; when it throws `AgentServerUnavailableError`, React Query does not retry before the root renders the recovery screen.
- Backend health probes use a 4s timeout and `retry: false`; one transient probe failure is recorded immediately and can make a backend appear disconnected even if it recovers moments later.

## Expected behavior

Transient network, tunnel, or timeout failures should retry briefly and show a reconnecting state. Non-transient failures, such as missing/invalid API keys or unsupported/unknown agent-server versions, should still fail fast and show actionable backend management errors.

## Proposed fix

- Retry bootstrap `/server_info` failures for a short bounded window when the backend is configured but temporarily unavailable.
- Show a reconnecting loading state during those retries instead of immediately opening backend recovery.
- Retry backend health probes a few times before recording a failure, while preserving fail-fast behavior for auth and compatibility errors.
- Add focused tests for retry classification and retry exhaustion behavior.

mvanhorn:
HUMAN:

AGENT:

Verified end-to-end against the change by building i18n and running the relevant Vitest suites and the type/lint gates locally in a fresh `npm ci` workspace:

```
$ npm run make-i18n
$ npx vitest run __tests__/hooks/query/use-backends-health.test.tsx __tests__/root.test.tsx
 Test Files  2 passed (2)
      Tests  32 passed (32)
$ npm run typecheck          # react-router typegen && tsc -> clean
$ npx eslint src             # exit 0
$ npx prettier --check <changed files>   # all matched files use Prettier code style
```

The new tests exercise the actual runtime paths: a transient `/server_info` failure (MSW `HttpResponse.error()` on the first request) now retries and renders the routed app instead of the recovery modal, while a 401 / unsupported-version still fails fast into the picker. The health-probe tests assert that a single transient probe failure recovers with `consecutiveFailures: 0` (no recorded failure), an exhausted transient budget records exactly one failure, and an invalid-API-key 401 records immediately without consuming the retry budget.

---

## Why

Agent Canvas drops the user into the backend recovery / picker UI after a single transient backend failure during startup, which is disruptive when the agent-server or tunnel is only briefly slow. Two startup paths are too sensitive: the bootstrap `/server_info` config query did not retry on `AgentServerUnavailableError` before `root.tsx` rendered the recovery screen, and backend health probes recorded a failure on the very first probe error, so one transient probe could mark a backend disconnected even when it recovered moments later. Non-transient failures (missing/invalid API key, unsupported agent-server version, auth) must still fail fast.

The subtlety: `AgentServerUnsupportedVersionError` and `AgentServerUnknownVersionError` both extend `AgentServerUnavailableError`, so a naive "retry on unavailable" would incorrectly retry real version-floor failures. Those, plus auth errors and the no-backend-configured sentinel, must keep failing fast.

## Summary

- Add `isTransientAgentServerError(error)` in `agent-server-compatibility.ts` that classifies a failure as transient only when it is an `AgentServerUnavailableError` that is not a version error, not an auth error, and not the no-backend-configured sentinel.
- `use-config.ts`: retry the bootstrap config probe a bounded number of times (3) on transient errors with a short capped delay; auth / version / no-backend errors fail fast immediately.
- `use-backends-health.ts`: add a bounded in-probe retry (2 extra attempts) for transient probe failures before calling `recordBackendFailure`; terminal failures record immediately, and `recordBackendFailure` is still called exactly once per failed query (preserving `MAX_CONSECUTIVE_FAILURES` semantics).

## Issue Number

Fixes #1420

## How to Test

```
npm ci
npm run make-i18n
npx vitest run __tests__/hooks/query/use-backends-health.test.tsx __tests__/root.test.tsx
npm run typecheck
npx eslint src
```

All targeted tests pass (32) and typecheck / lint are clean. The new test cases cover both the transient-retry recovery path and the fast-fail paths for auth and unsupported-version errors.

## Video/Screenshots

No UI surface changed; the behavior is verified through the added retry-classification unit/integration tests (terminal output above). The recovery modal and reconnecting paths render the same components as before — only the retry timing before the modal appears changed.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No migrations or config changes. The retry windows are small and bounded (config: 3 attempts with a capped delay; health probe: 2 extra attempts with a 250ms spacing), so a genuinely down backend still surfaces the picker quickly.

### DIFF
--- a/__tests__/hooks/query/use-backends-health.test.tsx
+++ b/__tests__/hooks/query/use-backends-health.test.tsx
@@ -119,9 +119,69 @@ describe("useBackendsHealth", () => {
       wrapper,
     });
 
-    await waitFor(() =>
-      expect(result.current[localBackend.id].isConnected).toBe(false),
+    await waitFor(
+      () => expect(result.current[localBackend.id].isConnected).toBe(false),
+      { timeout: 2000 },
     );
+  });
+
+  it("retries a transient probe failure before recording a backend failure", async () => {
+    getSettingsMock
+      .mockRejectedValueOnce(new Error("ECONNREFUSED"))
+      .mockResolvedValue({});
+
+    const { result } = renderHook(() => useBackendsHealth([localBackend]), {
+      wrapper,
+    });
+
+    await waitFor(
+      () =>
+        expect(result.current[localBackend.id]).toMatchObject({
+          isConnected: true,
+          consecutiveFailures: 0,
+        }),
+      { timeout: 1500 },
+    );
+    expect(getSettingsMock).toHaveBeenCalledTimes(2);
+  });
+
+  it("records a failure exactly once after exhausting the transient retry budget", async () => {
+    getSettingsMock.mockRejectedValue(new Error("ECONNREFUSED"));
+
+    const { result } = renderHook(() => useBackendsHealth([localBackend]), {
+      wrapper,
+    });
+
+    await waitFor(
+      () =>
+        expect(result.current[localBackend.id]).toMatchObject({
+          isConnected: false,
+          consecutiveFailures: 1,
+        }),
+      { timeout: 2000 },
+    );
+  });
+
+  it("records an invalid API key failure immediately without consuming the transient retry budget", async () => {
+    getSettingsMock.mockRejectedValue(
+      Object.assign(new Error("Unauthorized"), {
+        name: "HttpError",
+        status: 401,
+      }),
+    );
+
+    const { result } = renderHook(() => useBackendsHealth([localBackend]), {
+      wrapper,
+    });
+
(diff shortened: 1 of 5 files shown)


neubig:

PR #1139, opened 2026-06-04
fix: validate onboarding model and default per agent (#1111)

chuckbutkus:
## Summary

Fixes #1111 — first-run onboarding kept the OpenHands provider's default model `claude-opus-4-5`, which is an alias that not every API key is entitled to. The submit test then failed with:

```
litellm.BadRequestError: Litellm_proxyException - /chat/completions: Invalid model name passed in model=claude-opus-4-5. Call `/v1/models` to view available models for your key.
```

This PR resolves the issue with a focused, three-pronged fix that mirrors the resolution options listed in the issue:

1. **Pick a model that exists for the selected agent.** The onboarding LLM step now derives its initial `llm.model` from the agent chosen in step 1, so OpenHands defaults to a concrete Anthropic model instead of the OpenHands alias, and Claude Code / Codex / Gemini each get a sensible default.
2. **Validate the model against the key before persisting.** New `resolveOpenHandsModelForApiKey` calls `/api/llm/models` (with the user's provider + key + optional base URL) and resolves the selected model to the closest variant the key actually supports, falling back to a compatible variant when the literal alias is missing. The onboarding profile is created/activated with that resolved model rather than the raw default.
3. **Better empty-state UX.** The model selector now shows a `Select Model` placeholder when the field is empty and is hardened against browser autofill so onboarding stays deterministic.

This is a minimal extract from #1136 — only the four commits that address #1111. The unrelated websocket-recovery and version-bump commits from that PR are deliberately excluded so this can be merged without conflicts.

## Changes

- `src/utils/resolve-openhands-model.ts` (new) — provider/key-aware model resolution + fallback picker.
- `__tests__/utils/resolve-openhands-model.test.ts` (new) — unit coverage for the resolver.
- `src/components/features/onboarding/onboarding-modal.tsx` — pass the selected agent id into `SetupLlmStep`.
- `src/components/features/onboarding/steps/setup-llm-step.tsx` — per-agent default model map and resolve-before-save logic.
- `src/components/shared/modals/settings/model-selector.tsx` — `Select Model` placeholder and autofill hardening.
- `__tests__/components/onboarding/onboarding-modal.test.tsx` and `__tests__/components/modals/settings/model-selector.test.tsx` — updated coverage.
- `src/i18n/translation.json` — new placeholder string.

## Testing

- `npm run typecheck` ✅
[...]

PR #1518, opened 2026-06-27
fix: resolve OpenHands model alias against the API key when saving an LLM profile (#1111)

jpshackelford:
## Summary

This is a **follow-up stacked on top of #1139** (base: `fix/onboarding-model-validation`) that closes the remaining gap for issue **#1111**.

#1139 fixes the **first-run onboarding** path. But the same `Invalid model name` failure happens on the **manual LLM-profile path** — Settings → LLM Profiles → create/edit a profile, pick the **OpenHands** provider + a verified model (e.g. Sonnet 4.5). That flow (`llm-settings-local-view.tsx::handleSave`) persisted the bare alias verbatim, so the first conversation failed at runtime with:

```
litellm.BadRequestError: Litellm_proxyException - /chat/completions: Invalid model name passed in model=claude-sonnet-4-5. Call `/v1/models` to view available models for your key.
```

Root-cause trace and reproduction are documented in the issue: https://github.com/OpenHands/agent-canvas/issues/1111#issuecomment-4821426244

## What this PR adds (the delta on top of #1139)

1. **Apply resolution to the manual profile-save path.** `handleSave` now runs `resolveOpenHandsModelForApiKey` (the helper #1139 introduced for onboarding) before persisting, so a model that the key can't call is resolved to a callable catalog id instead of being saved as-is.
   - Resolution runs **only when a fresh API key is entered** (new profile, or the key is replaced). In edit mode the preserved `api_key` is the agent-server's encrypted token, which can't be used as a bearer credential — so we skip resolution when the key is unchanged (no wasted `/v1/models` round-trip).

2. **Harden `chooseOpenHandsFallbackModel` for provider-prefixed catalog ids.** The previous matcher only handled exact / case-insensitive / date-suffixed (`<model>-<date>`) variants. A key's `/v1/models` commonly lists **underlying-provider-prefixed** ids (e.g. `anthropic/claude-sonnet-4-5`) while the dropdown offers the bare name (`claude-sonnet-4-5`); none of the old branches matched those, so the resolver would fall through and persist the unusable alias. The matcher now also matches on the trailing path segment (exact and date-suffixed), e.g. `claude-sonnet-4-5` → `anthropic/claude-sonnet-4-5`.

   > ⚠️ @chuckbutkus / @FraterCCCLXIII — worth confirming the **public** proxy's `/v1/models` id shape. On a real LiteLLM proxy I inspected, the catalog is `anthropic/…`-prefixed, which the original matcher in #1139 would not resolve. This hardening covers both bare and prefixed shapes.

## Tests

- `__tests__/utils/resolve-openhands-model.test.ts` — new cases for provider-prefixed resolution (exact, latest-dated, and "prefer exact bare over prefixed").
- `__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx` — a real end-to-end (MSW) test that entering a key on a profile with a bare OpenHands alias queries `/v1/models` with the typed bearer token and persists the resolved `openhands/anthropic/claude-sonnet-4-5`; plus a negative test that no `/v1/models` call happens when the key is unchanged.

All 24 tests in the touched component suite pass; `npm run typecheck` is clean. (Note: the test file has 3 pre-existing `prettier/prettier` lint errors inherited from the base branch, unrelated to this change.)

## Suggested handling

Since this is stacked on your branch, the cleanest path is to **merge/cherry-pick these commits into #1139** so #1111 is fully fixed in one PR. Happy to adjust the approach (e.g. retarget to `main`, or fold the matcher hardening directly into your `resolve-openhands-model.ts`) — whatever's easiest for you.

Refs #1111. Complements #1139.

---
_This PR was created by an AI agent (OpenHands) on behalf of @jpshackelford._

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/6f29e4de-4dcb-4163-a8b5-50cb51b06d6e)

### DIFF
--- a/__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx
+++ b/__tests__/components/settings/llm-profiles/llm-settings-local-view.test.tsx
@@ -1,8 +1,10 @@
 import { describe, expect, it, vi, beforeEach, type Mock } from "vitest";
+import { http, HttpResponse } from "msw";
 import { AxiosError } from "axios";
 import { screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { renderWithProviders } from "test-utils";
+import { server } from "#/mocks/node";
 import {
   LlmSettingsLocalView,
   shouldReapplyProfileAfterSave,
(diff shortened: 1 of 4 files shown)


neubig:

PR #1532, opened 2026-06-29
feat: support LLM profiles on cloud backends

### LINKED ISSUE #533: Tech debt: unify local and cloud LLM configuration around profiles
## Summary
Today the UI has to special-case local vs cloud LLM configuration:
- local `/settings` is profile-oriented (`LLM Profiles`)
- cloud `/settings` is still raw `LLM Settings`
- footer/menu wording also has to branch by backend mode

This is functionally correct, but it creates product and implementation friction. The same conceptual surface (choosing/configuring how the agent uses an LLM) presents itself differently depending on backend kind.

## Problem
We currently expose two mental models:
1. **Local**: saved LLM profiles that can be created, edited, and activated.
2. **Cloud**: direct raw LLM settings.

That leaks backend details into the product UX and forces frontend copy / navigation / settings surfaces to diverge.

## Goal
Find a way to present **one user-facing profile model** across local and cloud, even if the underlying storage/mechanics differ.

## Possible direction
One possible approach:
- keep **profiles as the UI abstraction** everywhere
- on cloud, store profiles client-side (or in a lightweight app-layer store)
- when starting a conversation or saving cloud settings, **materialize/flatten** the selected profile into the raw `llm` payload expected by the cloud path

This would let us:
- avoid backend-specific copy like `LLM Profiles` vs `LLM Settings`
- keep footer/settings navigation semantics consistent
- let users think in terms of named configurations regardless of backend kind

## Open questions
- Where should cloud-side profile data live (browser localStorage, app backend, agent-server proxy, etc.)?
- Should cloud support exactly one effective profile at a time, or multiple named saved profiles like local?
- How should profile secrets/API keys be stored and redacted safely on cloud?
- Should profile switching affect only new conversations, or also live conversations when supported?
- Can we converge the local and cloud settings APIs enough that the UI no longer branches on backend kind?

## Why now
Recent UI work exposed the awkwardness here directly:
- the local settings page needed to be renamed to `LLM Profiles`
- local footer links needed profile wording
- cloud still has to keep `LLM Settings`

That is a reasonable short-term fix, but it is a sign that the conceptual model should probably be unified.

VascoSch92:
HUMAN:

support LLM profiles on cloud backends

AGENT:

Evidence this runs end-to-end:

- **Tests** — 238 tests pass across the touched surface (`npx vitest run`), including the new `__tests__/api/cloud/profiles-service.test.ts` (cloud profile routing through `ProfilesService`), `__tests__/api/agent-server-conversation-service.test.ts` (asserts a cloud per-conversation switch issues `POST /api/v1/app-conversations/{id}/switch_profile` with `{profile_name}`), `__tests__/routes/llm-settings.test.tsx`, `__tests__/components/features/chat/components/chat-input-actions.test.tsx`, `__tests__/hooks/chat/use-slash-command.test.ts`, `__tests__/hooks/chat/use-model-interceptor.test.tsx`, plus the existing `__tests__/components/settings/llm-profiles/*` and `use-llm-configured` suites (local behavior unchanged).
- **Static checks** — `tsc` clean on all changed files; `eslint` and `prettier --check` clean on the 12 files. (Repo-wide `tsc` currently errors only on unrelated untracked `agents` WIP that is not part of this branch.)
- **App boots** — `npm run dev` serves the GUI (`GET http://localhost:3001` → `200`, `<title>OpenHands</title>`); the LLM settings route renders the profile manager and the chat composer renders the profile switcher.
- **Left for human verification** — a full click-through against a *live* cloud backend (create/activate/switch a profile end-to-end through a running app-server) needs cloud credentials; see the HUMAN checkbox.

---

## Why

Cloud backends had no access to LLM profiles. The LLM was configured only through the flat cloud settings form, and the chat composer showed a plain model picker — even though the cloud app-server already exposes the full profile machinery under `/api/v1/settings/profiles` and a server-resolved per-conversation `/api/v1/app-conversations/{id}/switch_profile`. This brings cloud to parity with local.

## Summary

- **Data layer:** new `src/api/cloud/profiles-service.api.ts`; `ProfilesService` branches to it when the active backend is cloud (mirrors `SettingsService` → `fetchCloudSettings`), so the existing profile hooks and settings manager work transparently.
- **Settings page:** the LLM settings route renders the profile manager for both backends (was cloud → plain form).
- **Chat switching:** the composer shows the profile switcher on cloud, `/model` lists/switches profiles, and per-conversation switching routes through the app-server's server-resolved `switch_profile` endpoint.

## Issue Number

Fix https://github.com/OpenHands/agent-canvas/issues/533

## How to Test

1. `npm install && npm run dev`, open http://localhost:3001.
2. Connect to a cloud backend (Manage Backends) and select it.
3. **Settings → LLM**: you should see the profile manager (Add LLM Profile + a list with activate / edit / rename / duplicate / delete) instead of the flat form. Create a profile, activate it, then rename/delete it.
4. **Chat composer**: open a conversation on the cloud backend — the profile switcher pill appears; click it to switch. Try `/model` (lists profiles inline) and `/model <name>` (switches). Each per-conversation switch fires `POST /api/v1/app-conversations/{id}/switch_profile`.
5. **Local backends**: confirm no regression — the manager, switcher, and `/model` behave as before.

## Video/Screenshots

https://github.com/user-attachments/assets/5cf79f5f-947e-462d-9f16-550a035ec1b1

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Companion PR:** OpenHands/OpenHands#15021 — cloud conversations don't currently enable LLM token streaming (the app-server builds the agent/profile LLM with `stream=False`, the SDK default), so a cloud profile switch won't stream until that ~2-line app-server fix forces `stream=True`.
- Duplicating a BYOR profile on cloud starts keyless (the cloud never echoes `api_key`). A cloud user with zero saved profiles sees no composer model control (the switcher self-hides) — identical to local.
- `use-ensure-active-profile` (auto-activation) is intentionally left local-only.

---

| Component | Value |
|---|---|
| **Agent Server** | `ghcr.io/openhands/agent-server:1.29.3-python` |
| **Automation** | `openhands-automation==1.0.0a13` |
| **Commit** | `1c22c2d57d4c8de1af0dda408cf54f234b73e8f2` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-1c22c2d
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-1c22c2d
```

```
ghcr.io/openhands/agent-canvas:sha-1c22c2d-amd64
ghcr.io/openhands/agent-canvas:vasco-profiles-and-be-amd64
ghcr.io/openhands/agent-canvas:pr-1532-amd64
ghcr.io/openhands/agent-canvas:sha-1c22c2d-arm64
ghcr.io/openhands/agent-canvas:vasco-profiles-and-be-arm64
ghcr.io/openhands/agent-canvas:pr-1532-arm64
ghcr.io/openhands/agent-canvas:sha-1c22c2d
ghcr.io/openhands/agent-canvas:vasco-profiles-and-be
ghcr.io/openhands/agent-canvas:pr-1532
```

- Each tag (e.g., `sha-1c22c2d`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-1c22c2d-amd64`) are also available if needed

### DIFF
--- a/src/hooks/use-can-manage-llm-profiles.ts
+++ b/src/hooks/use-can-manage-llm-profiles.ts
@@ -1,0 +1,39 @@
+import { useQuery } from "@tanstack/react-query";
+import { getCloudOrganizationMe } from "#/api/cloud/organization-service.api";
+import { useActiveBackend } from "#/contexts/active-backend-context";
+
+/**
+ * Whether the current user may MUTATE LLM profiles on the active backend —
+ * create, edit, rename, delete, duplicate, or activate/switch.
+ *
+ * - Local agent-server (OSS): always `true`; the user owns their own profiles.
+ * - Cloud: profiles are org-scoped. The app-server grants every mutating
+ *   profile action (save/delete/rename/activate) only to the `owner` and
+ *   `admin` roles via `EDIT_ORG_SETTINGS`; a `member` has `VIEW_ORG_SETTINGS`
+ *   and is view-only. We read the caller's role from
+ *   `GET /api/organizations/{orgId}/me` — the same call `useCloudCurrentUserId`
+ *   makes, reusing its query key so no extra request is issued.
+ *
+ * Returns `false` while the role is still loading or unknown on cloud, so
+ * mutating controls never flash for a member before the role resolves.
+ */
+export function useCanManageLlmProfiles(): boolean {
+  const { backend, orgId } = useActiveBackend();
+  const isCloud = backend.kind === "cloud";
+
+  // `backend` is identified by `backend.id`, which is already in the key; we
+  // keep the key byte-identical to useCloudCurrentUserId so React Query shares
+  // the cached /me result instead of firing a second request.
+  // eslint-disable-next-line @tanstack/query/exhaustive-deps
+  const { data } = useQuery({
+    queryKey: ["cloud-current-user", backend.id, orgId],
+    queryFn: () => getCloudOrganizationMe(orgId!, backend),
+    enabled: isCloud && !!orgId,
+    staleTime: 1000 * 60 * 5,
+    retry: false,
+    meta: { disableToast: true },
+  });
+
+  if (!isCloud) return true;
+  return data?.role === "owner" || data?.role === "admin";
+}
--- a/src/hooks/use-can-manage-llm-profiles.ts
+++ b/src/hooks/use-can-manage-llm-profiles.ts
@@ -3,19 +3,27 @@ import { getCloudOrganizationMe } from "#/api/cloud/organization-service.api";
 import { useActiveBackend } from "#/contexts/active-backend-context";
 
 /**
+ * Server-defined permission (mirrors the app-server's
+ * `authorization.Permission.EDIT_ORG_SETTINGS`) that gates LLM-profile
+ * mutations for an org.
+ */
+const EDIT_ORG_SETTINGS = "edit_org_settings";
+
+/**
  * Whether the current user may MUTATE LLM profiles on the active backend —
  * create, edit, rename, delete, duplicate, or activate/switch.
  *
  * - Local agent-server (OSS): always `true`; the user owns their own profiles.
  * - Cloud: profiles are org-scoped. The app-server grants every mutating
- *   profile action (save/delete/rename/activate) only to the `owner` and
- *   `admin` roles via `EDIT_ORG_SETTINGS`; a `member` has `VIEW_ORG_SETTINGS`
- *   and is view-only. We read the caller's role from
+ *   profile action (save/delete/rename/activate) only to the `owner`/`admin`
+ *   roles via the `edit_org_settings` permission; a `member` is view-only. We
+ *   read the caller's server-defined permissions from
  *   `GET /api/organizations/{orgId}/me` — the same call `useCloudCurrentUserId`
- *   makes, reusing its query key so no extra request is issued.
+ *   makes, reusing its query key so no extra request is issued — and fall back
+ *   to a role check when an older app-server doesn't return `permissions`.
  *
- * Returns `false` while the role is still loading or unknown on cloud, so
- * mutating controls never flash for a member before the role resolves.
+ * Returns `false` while the role/permissions are still loading or unknown on
+ * cloud, so mutating controls never flash for a member before they resolve.
  */
 export function useCanManageLlmProfiles(): boolean {
   const { backend, orgId } = useActiveBackend();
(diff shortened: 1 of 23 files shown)


hieptl (changes requested):
Hello @VascoSch92,

While testing this pull request, I noticed an issue that I wanted to bring to your attention.

In OpenHands Cloud, I'm a member of an organization called "Demo Org," where my role is **Member**. As expected, I don't have permission to modify LLM profiles and can only view them.

However, from the agent-canvas, I can see several options for interacting with LLM profiles, and I'm also able to edit Cloud LLM profiles. This seems inconsistent with the permissions assigned to my role.

I've attached a video below that demonstrates the behavior for reference.

https://github.com/user-attachments/assets/76b7f94f-a67e-45e1-a5eb-b16e259152bc

Thank you very much! 🙏

kripper:
Nice. The lack of multiple LLM profiles was stopping me from using agent-canvas.
BTW, I noticed that when you remove a profile from the old OH frontend, the agent-server continues using that deleted profile for the condenser or for the /btw messages,
and I also noticed an issue with propagating the LLM timeout settings to the condenser LLM,
it's like agent-server continues using the LLM settings that were available when the conversation was created.

VascoSch92:
Hey @hieptl 

thanks for the review. 

I updated the code.

However, I cannot test it as I don't have access to org. Could you add me?

VascoSch92:
https://github.com/user-attachments/assets/7ebe2718-98e2-4b82-8d98-fdb2b982a862

src/hooks/use-can-manage-llm-profiles.ts
```diff
@@ -0,0 +1,39 @@
...
+    queryKey: ["cloud-current-user", backend.id, orgId],
+    queryFn: () => getCloudOrganizationMe(orgId!, backend),
+    enabled: isCloud && !!orgId,
+    staleTime: 1000 * 60 * 5,
+    retry: false,
+    meta: { disableToast: true },
+  });
+
+  if (!isCloud) return true;
+  return data?.role === "owner" || data?.role === "admin";
```
hieptl, line 52: **suggestion (blocking):** I think it would be better to define this permission on the server side rather than comparing roles on the client side. You can refer to the `authorization.py` file in the OpenHands repository for an example of how permissions are handled.

In addition, it would be good to validate permissions on both the client and the server. At the moment, the validation only happens on the client side. This means that anyone who knows the API key and calls the APIs directly could bypass the client-side check.

Thank you very much! 🙏

hieptl (changes requested):

